PR #1, opened 2019-11-30
Documentation builds

ammen99:
This is the basic framework for building the documentation from source code.
I have also started working on some of the classes in the branch [job](https://github.com/DistributedTaskScheduling/JobAdder/tree/job) which can be used for testing these.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,6 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+
+# JobAdder-specific
+./build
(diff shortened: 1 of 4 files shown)


ammen99:
Forgot to mention something, I separated the Doxyfile and py-puml-tools into a separate repository, since these files are big so it will then make the statistics about lines added/removed quite unbalanced.

ammen99:
> I'm not sure how you are imagining the final workflow.
> This PR would add a submodule of the form JobAdder -> doctools.
> Are we going to later add JobAdder as a submodule to Documentation?

You have a point here.
The problem we have: we need all repositories to be able to build the final PDF from the documentation phase (parts of it are code, parts are sequence diagrams, associations, etc.).
Maybe we should put design into its own repository and add it as a submodule to JobAdder (kind-of the docs/ folder in this PR)

PR #2, opened 2019-11-30
Description of Job class

ammen99:
I started working on https://github.com/DistributedTaskScheduling/documentation/issues/64

### DIFF
--- a/src/ja/common/docker_context.py
+++ b/src/ja/common/docker_context.py
@@ -1,0 +1,52 @@
+'''
+Contains definitions of the docker context of a job.
+'''
+from abc import ABC, abstractmethod
+from typing import List
+
+
+class MountPoint:
+    '''
+    A mount point consists of:
+    1. A directory to be mounted.
+    2. The target path in the Docker container where the directory needs to be
+    mounted.
+    '''
+    def __init__(self, source: str, target: str):
+        '''
+        Initializes a new mount point.
+        @source The directory to be mounted.
+        @target The target path in the container to mount at.
+        '''
+
+    @property
+    def source_path(self) -> str:
+        '''
+        @return The directory to be mounted.
+        '''
+
+    @property
+    def mount_path(self) -> str:
+        '''
+        @return The path where the directory should be mounted.
+        '''
+
+
+class IDockerContext(ABC):
+    '''
+    A docker context consists of the necessary data to build a docker image to
+    run a job in.
+    '''
+
+    @abstractmethod
+    def get_dockerfile_source(self) -> str:
+        '''
+        @return The string contents of a Dockerfile which can be used to build
+        the docker image.
+        '''
+
+    @abstractmethod
+    def get_mount_points(self) -> List[MountPoint]:
+        '''
+        @return A list of all mount points for the job.
+        '''
--- a/src/ja/common/docker_context.py
+++ b/src/ja/common/docker_context.py
@@ -1,88 +1,88 @@
-'''
+"""
 Contains definitions of the various Docker-related attributes of a job.
 These include, but are not limited to, mount points, program environment
 and hardware constraints.
-'''
+"""
 from abc import ABC, abstractmethod
 from typing import List
 
 
 class MountPoint:
-    '''
+    """!
     A mount point consists of:
     1. A directory to be mounted(@source_path).
     2. The target path in the Docker container where the directory needs to be
     mounted(@mount_path).
-    '''
+    """
     def __init__(self, source_path: str, mount_path: str):
-        '''
+        """!
         Initializes a new mount point.
         @param source_path The host directory to be mounted.
         @param mount_path The target path in the container to mount at.
-        '''
+        """
 
     @property
     def source_path(self) -> str:
-        '''
+        """!
         @return The directory to be mounted.
-        '''
+        """
 
     @property
     def mount_path(self) -> str:
-        '''
+        """!
         @return The path where the directory should be mounted.
-        '''
+        """
 
 
 class IDockerContext(ABC):
-    '''
+    """
     A docker context consists of the necessary data to build a docker image to
     run a job in.
-    '''
+    """
 
     @abstractmethod
     def get_dockerfile_source(self) -> str:
-        '''
+        """!
         @return The string contents of a Dockerfile which can be used to build
         the docker image.
-        '''
+        """
 
     @abstractmethod
     def get_mount_points(self) -> List[MountPoint]:
-        '''
+        """!
         @return A list of all mount points for the job.
-        '''
+        """
 
 
 class DockerConstraints:
-    '''
+    """
     A list of constraints of the docker container.
-    '''
+    """
     def __init__(self, cpu_threads: int = -1, memory: int = 1):
-        '''
+        """!
         Create a new set of Docker constraints.
         @param cpu_threads Initial value of @cpu_threads.
         @param memory The value of @memory.
-        '''
+        """
 
     @property
     def cpu_threads(self) -> int:
-        '''
-        The maximum number of CPU threads the docker container can have.
+        """!
+        @return The maximum number of CPU threads the docker container can use.
         -1 means that the amount of threads is not set, in which case the
         number of threads will be determined when scheduling the job.
-        '''
+        """
 
     @cpu_threads.setter
     def cpu_threads(self, count_threads: int) -> None:
-        '''
+        """!
         Set the number of CPU threads.
         If the number of cpu threads is already set(i.e it is not equal to -1),
         this will raise a RuntimeError.
-        '''
+        """
 
     @property
     def memory(self) -> int:
-        '''
-        The maximum amount of RAM in MB to allocate for this container.
-        '''
+        """!
+        @return The maximum amount of RAM in MB to allocate for this container.
+        """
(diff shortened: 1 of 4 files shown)


src/ja/common/job.py
```diff
@@ -0,0 +1,103 @@
...
+
+class JobStatus(Enum):
+    '''
+    Represents the status of a job.
+    '''
+    NEW = 0
+    QUEUED = 1
+    RUNNING = 2
+    PAUSED = 3
+    DONE = 4
```
JohannesGaessler: Are we always going to instantly remove the job from the worker client when it's done?

src/ja/common/job.py
```diff
@@ -0,0 +1,103 @@
...
+    KILLED = 5
+    CRASHED = 6
+
+
+class JobPriority(Enum):
+    '''
+    Represents the priority of a job.
```
JohannesGaessler: priority -> base priority

src/ja/common/job.py
```diff
@@ -0,0 +1,103 @@
...
+    HIGH = 2
+    URGENT = 3
+
+
+class Job:
+    '''
+    Represents a job with all of its attributes.
+    '''
+    def __init__(self,
+                 owner_id: int,
```
JohannesGaessler: I think owner_id should not be in the constructor.
The way I think our project will work:

1. User constructs Job object on user client.
2. Job object is turned into YAML.
3. YAML is transferred to central server.
4. Job object is reconstructed from YAML.
5. Central server validates Job object and **modifies it**.

I think one of the modifications in step 5 will be to set the owner_id.
As such I think that the Job constructor should not contain it because it will only be assigned after object creation.

src/ja/common/job.py
```diff
@@ -0,0 +1,103 @@
...
+                 priority: JobPriority,
+                 ctx: IDockerContext,
+                 constraints: DockerConstraints):
```
JohannesGaessler: Add kwarg for human readable Job name.

src/ja/common/job.py
```diff
@@ -0,0 +1,103 @@
...
+        self._uid = None
+        self._status = JobStatus.NEW
+        self._priority = priority
+        self._ownerid = owner_id
+        self._ctx = ctx
+        self._constraints = constraints
+
+    @property
+    def uid(self) -> str:
```
JohannesGaessler: Should this be str or int?
If a user later queries the server is this going to be a visible property?

src/ja/common/job.py
```diff
@@ -0,0 +1,103 @@
...
+        '''
+
+    @property
+    def owner_id(self) -> int:
+        '''
+        Get the id of the user who owns this job.
+        '''
+
+    @property
+    def priority(self):
```
JohannesGaessler: Type hint.

src/ja/common/job.py
```diff
@@ -0,0 +1,103 @@
...
+        '''
+        Get the job priority.
+        '''
+
+    @property
+    def status(self):
```
JohannesGaessler: Type hint.

src/ja/common/job.py
```diff
@@ -0,0 +1,103 @@
...
+    def status(self, new_status: JobStatus) -> None:
+        '''
+        Set the job status.
+        The following transitions are allowed:
+        NEW -> QUEUED
+        QUEUED -> KILLED | RUNNING
+        RUNNING -> PAUSED | DONE | KILLED | CRASHED
+        PAUSED -> KILLED | RUNNING
+
+        If the transition requested is not allowed, or if the job has notb
```
JohannesGaessler: notb -> not

src/ja/common/docker_context.py
```diff
@@ -0,0 +1,85 @@
...
+        -1 means that the amount of threads is not set, in which case the
+        number of threads will be determined when scheduling the job.
+        '''
+
+    @cpu_threads.setter
+    def cpu_threads(self, count_threads: int) -> None:
+        '''
+        Set the number of CPU threads.
+        If the number of cpu threads is already set(i.e it is not equal to -1),
+        this will raise a RuntimeError.
```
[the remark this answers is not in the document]
ammen99: Yes, I think thread count is set when you queue the job?

src/ja/common/job.py
ammen99: What do you mean by remove?

src/ja/common/job.py
ammen99: I was thinking that the job is created with the userid by the user client and then the server just checks whether the user communicating with the server is the same one as the user in the job.

src/ja/common/docker_context.py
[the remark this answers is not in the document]
JohannesGaessler: I'm asking because we did not specify modification of pending jobs in the requirements document.
I'm also not sure if we talked about it in the meeting.
I think to keep things simple we should not allow modification of pending jobs.
Users should instead kill pending jobs and re-add them.

src/ja/common/job.py
ammen99: In our requirements we defined UID as a string consisting of letters, digits and -
And yes it will be visible, otherwise the user can't say which of their jobs to cancel (or which job to query information about)

src/ja/common/job.py
JohannesGaessler: I was thinking about SSH into a container that is done and if we would need an extra state to differentiate if the container still being loaded on the machine.
But I think we decided in the meeting that users should only be allowed to SSH into running containers.
Sorry, my bad.

src/ja/common/job.py
JohannesGaessler: You're right.
That would work just as well.

src/ja/common/job.py
JohannesGaessler: I see.
It seems I got confused.
Should UID be part of the constructor?

ammen99: I was initially thinking it should be in the constructor, but after all, we can't get a UID for the job in the user client - the server assigns UIDs (otherwise we *might* have collisions). So a job initially has no UID, and when it was Queued, the server should respond back with a UID.

However, we may want to take a different approach. I am also not sure where this piece of information belongs. Should I add it in the comment here?

JohannesGaessler: I think it should work like this:

* UID is None by default. In this case the server automatically assigns a UID when it receives the Job.
* A user can specify a desired UID when creating the Job.
* If the server detects a collision it renames the Job (something like Tensorflow -> Tensorflow-2).

src/ja/common/docker_context.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class DockerConstraints:
+    '''
+    A list of constraints of the docker container.
+    '''
+    def __init__(self, count_threads: int = -1, memory: int = 1):
+        '''
+        Create a new set of Docker constraints.
+        memory is measured in MB.
```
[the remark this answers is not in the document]
ammen99: It is explained in the property, do I need to repeat it here?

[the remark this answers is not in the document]
ammen99: Maybe I can just say that count_thread (after renaming, cpu_threads) is the initial value for the attribute.

ammen99:
Still TODO:

- [ ] Add list of shared resources (the license things) to a job.

- [ ] Add non-preemptible flag to a job.

src/ja/common/docker_context.py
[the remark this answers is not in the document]
JohannesGaessler: > do I need to repeat it here

I think all arguments should be documented, even if it's repetitive.
This is because of how the documentation will be used.
Imagine someone who wants to know what arguments the Job constructor takes.
They will typically use their IDE to jump to the constructor definition.
If the constructor does not explain the arguments they would get confused.
Also consider that IDEs typically only process the immediate documentation of a method/constructor to assist the user.

src/ja/common/job.py
```diff
@@ -0,0 +1,137 @@
...
+    """
+    Represents the base priority of a job.
+    """
+    LOW = 0
+    MEDIUM = 1
+    HIGH = 2
+    URGENT = 3
+
+
+class JobSchedulingConstraints:
```
JohannesGaessler: How are you deciding which parts of the Job class to put in this class?
It seems kind of arbitrary.
I could argue that a Job's cpu_threads and memory affect "how it is scheduled".

ammen99: I split them by what the constraints apply to. RAM and CPU are constraints for the work machine, whereas priority, preemption, etc. are related purely to the scheduling.

The reason why I split them in two parts is purely aesthetic: I am not a fan of constructors or any method which take a lot of paramaters.


I could merge them together though if you are against the split.

nikolatzotchev:
Should we put the job stuff in a package job, in our current common package there is _message_ and _proxy_. I'm not entirely sure where we should put the other stuff (docker context).

src/ja/common/docker_context.py
```diff
@@ -0,0 +1,88 @@
...
+    A mount point consists of:
+    1. A directory to be mounted(@source_path).
+    2. The target path in the Docker container where the directory needs to be
+    mounted(@mount_path).
+    """
+    def __init__(self, source_path: str, mount_path: str):
+        """!
+        Initializes a new mount point.
+        @source_path The host directory to be mounted.
+        @mount_path The target path in the container to mount at.
```
JohannesGaessler: Shouldn't these be "@param source_path", "@param mount_path"?

ammen99: Yes, you're right.

ammen99:
I squashed some of the commits together and rebased on top of master.

PR #4, opened 2019-12-02
add .travis.yml and .travis/check_init_py.sh

### LINKED ISSUE #3: Decide on coding style and add CI
We really should have a consistent style. My proposal is to use the official Python styleguide: [PEP 8](https://www.python.org/dev/peps/pep-0008/).There are also good tools which check this, for example flake8.

Another thing is checking the type hints consistency with MyPy.

Anything else we want to check?

ammen99:
Set up a CI to check code style and type hints.

MyPy seems to require an `__init__.py` file in each source folder, even with the `--namespace-packages` option. So I also added a script to check that each of our directories in `src/` has an `__init__.py` file.

~~Marking this as a draft since we have not all spoken about and agreed on which Python coding style we're going to follow.~~

Fixes #3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: python
+python: 3.6
+
+install: "pip3 install flake8 mypy"
+script:
+  - flake8 src/
+  - ./.travis/check_init_py.sh src/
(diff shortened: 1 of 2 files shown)


ammen99:
Merging since changes are relatively trivial.

PR #5, opened 2019-12-04
issue #67

jooe1:
SchedulerLowLevel


nikolatzotchev:
I don't think python supports calling functions on a remote machine, that is why we discussed in discord how will that look like, so I think your communication with the WMClient is a bit off. Also I don't think it is necessary to wait for the urgent job to finish. It can be also another job.

ammen99:
I think there has been a misunderstanding. https://github.com/DistributedTaskScheduling/documentation/issues/67 is about the internal work of the scheduler. Worker client should not even be part of the sequence diagram.

For the diagram that you should make, you have the following parts:
1. Scheduler fetches job information + currently assigned jobs to work machines + work machine information from the database.

2. Scheduler decides which jobs should run on each machine (this is the most important part, here you may have multiple classes and methods)

3. Scheduler sends the new job distribution to the dispatcher.


The rest of the diagram is done in https://github.com/DistributedTaskScheduling/documentation/issues/66



Second: the sequence diagrams do not belong to this repository. Only python code, and associations (but associations we're going to work on during the next weeks).

Third: put all the code into a folder `src/ja/{server/common/worker/user_client}`. DO NOT USE `-` in the names, I do not think Python allows such a character in module names.

Fourth: we decided we will not develop methods yet, right? Also, you have just specified the type as `Job`, `Priority`, etc. They are not defined yet.


These are the problems with this PR as I see it. Please correct me if there is a misunderstanding about something on my part.

jooe1:
@nikolatzotchev The waiting is just to show that the paused job won't be resumed when the urgent job is still running
@ammen99 I think method stubs are better to describe classes (I am following the example provided by florian).

M1keReck:
Updated diagrams:
addSqD
CancelSqD

nikolatzotchev:
1. We decided it is going to be called WC not WMC.
2. I don't quite understand why you need RunningJobEntry, the only thing that it does is map a jobid to wcid.
3. Generally speaking i think it is better to make writings in the db after the thing is done not before that.
4. Shouldn't the authentication part happen before we are already in the Scheduler.
5. Class names should be starting with a capital letter, also we agreed on pep 8, the indentation should be 4 spaces and class names should start with a capital letter.
I will take a more detailed look later.

ammen99:
> @nikolatzotchev The waiting is just to show that the paused job won't be resumed when the urgent job is still running
> @ammen99 I think method stubs are better to describe classes (I am following the example provided by florian).

I think that either you have not been listening to what we discussed, or you simply forgot: We decided we're first going to do sequence diagrams and describe what each class is responsible for. And we **explicitly** said that methods will be done over the next two weeks.

Also @jooe1, you seem to think that naming and code style is not important: "because names and formatting are too important". We all agreed to follow PEP8. If you have objections, please tell the rest of the group what you don't like about it.

Now technical comments:

1. Authentication happens before the request is received in the server daemon, this is what my diagram is for (I will be uploading it shortly). You should not include it here.

2. You seem to be putting scheduling logic in the dispatcher. I think we agreed that the dispatcher simply generates a list of specific commands for the work machine, any scheduling (including which jobs are going to be paused, etc.) belongs to the scheduler.

src/ja/server/diagrams/schedulerLowLevel.puml
```diff
@@ -0,0 +1,129 @@
...
+    participant ":RunningJobEntry"
+    participant ":DataBaseHandler"
+    participant ":Dispatcher"
+end box
+
+activate "job:Job"
+-> ":QueuingActivity":
+activate ":QueuingActivity"
+
+":QueuingActivity" -> ":Scheduler" ++: queueJob(job:Job)
```
JohannesGaessler: Scheduler is supposed to retrieve list of Jobs from the Database.
When a Job is added that is merely a trigger for the Scheduler to retrieve a list of Jobs.

src/ja/server/diagrams/schedulerLowLevel.puml
```diff
@@ -0,0 +1,129 @@
...
+":Scheduler" -> ":DataBaseHandler" ++: getWCList()
+note left of ":Scheduler"
+    Scheduler decides which jobs
+    should run on each machine
+end note
+":DataBaseHandler" --> ":Scheduler" --: WCList:list[]
+":Scheduler" --> ":Scheduler" --: wmID:int
+
+":Scheduler" -> "job:Job" ++: set_wm_ID(wmID)
+"job:Job" --> ":Scheduler" --:
```
JohannesGaessler: Replace lines 26 to 36 with the following:
* Retrieve list of Jobs from Database
* Retrieve list of WCs from Database
* Sort Jobs by priority

For each Job:
* find_wc
* Set Job wm_id

src/ja/server/diagrams/schedulerLowLevel.puml
```diff
@@ -0,0 +1,129 @@
...
+end note    
+":RunningJobEntry" --> ":Scheduler" --: rje
+break   p==URGENT
+    ":Scheduler" -> ":DataBaseHandler" ++: addRunningJob(rje)
+    note right of ":DataBaseHandler" 
+        RunningJobEntry contains the
+        job ID and work machine ID
+    end note
+    ":DataBaseHandler" --> ":Scheduler" --:
+    ":Scheduler" -> ":Dispatcher" ++: dispatchUrgentJob(job)
```
JohannesGaessler: Dispatcher should be the one to change the sate of Jobs, not the Scheduler.

src/ja/server/diagrams/schedulerLowLevel.puml
```diff
@@ -0,0 +1,129 @@
...
+    ":Dispatcher" -> ":Dispatcher" ++: start(job)
+    ":Dispatcher" --> ":Dispatcher" --: message
+    ":Dispatcher" --> ":Scheduler": message
+    deactivate ":Dispatcher"
+    ":Scheduler" --> ":QueuingActivity": message
+    <-- ":QueuingActivity" : message
+
+end
+
+":Scheduler" -> "job:Job" ++: calcCost()
```
JohannesGaessler: Job cost functions must be evaluated **before** the Scheduler runs, not afterwards.

src/ja/server/diagrams/cancelLowLevel.puml
```diff
@@ -0,0 +1,55 @@
...
+    participant ":Authenticator"
+    participant ":DataBaseHandler"
+    participant ":Dispatcher"   
+end box
+
+
+-> ":CancelActivity":
+activate ":CancelActivity"
+
+":CancelActivity" -> ":Scheduler" ++: cancelJob(jobID:int, userID:int)
```
JohannesGaessler: The Scheduler should not be invoked to cancel a Job.
Cancel requests should be sent directly to the dispatcher.
The Scheduler should be invoked afterwards to re-evaluate which Jobs should be run on which machine.

src/ja/server/diagrams/cancelLowLevel.puml
```diff
@@ -0,0 +1,55 @@
...
+
+":Scheduler" -> ":Authenticator" ++:checkPermission(jobID, userID)
+":Authenticator" --> ":Scheduler" --: isPermitted:bool
+
+alt isPermitted
+    ":Scheduler" -> ":DataBaseHandler" ++: checkQueue(jobID)
+    ":DataBaseHandler" --> ":Scheduler" --:isQueued:bool
+
+    break isQueued
+        ":Scheduler" -> ":DataBaseHandler" ++: removeFromQueue(jobID)
```
JohannesGaessler: Set Job status in the Database to killed.

src/ja/server/diagrams/cancelLowLevel.puml
```diff
@@ -0,0 +1,55 @@
...
+        ":DataBaseHandler" --> ":Scheduler" --: 
+        ":Scheduler" -> ":Dispatcher" ++:  removeFromQueue(jobID)
```
JohannesGaessler: The Dispatcher does not have a queue of Jobs.
Remove this call.

src/ja/server/diagrams/cancelLowLevel.puml
```diff
@@ -0,0 +1,55 @@
...
+    ":DataBaseHandler" --> ":Scheduler" --
+    ":Scheduler" -> ":Dispatcher" ++: removeRunningJob(jobID)
+    ":Dispatcher" -> ":Dispatcher" ++: findJob(jobID) 
+    ":Dispatcher" --> ":Dispatcher" --: wmID:int
+
+    ":Dispatcher" -> ":Dispatcher" ++: stop(jobID, wmID)
+    ":Dispatcher" --> ":Dispatcher" --: 
+    ":Dispatcher" --> ":Scheduler" --: message
+    ":Scheduler" --> ":CancelActivity" --: message
+    <-- ":CancelActivity" : message
```
JohannesGaessler: Cancelling a Job that is already running should work like this:
* Retrieve Job from Database
* Tell the Dispatcher to kill the Job
* The Dispatcher get the id of the WorkerClient from the Job object.
* The Dispatcher instructs a WorkerClientProxy to kill the Job.
* If successful, update Database and re-evaluate which Jobs should be run on which WorkerClients. This re-evaluation is the **exact same** like when a Job is added by a user.

src/ja/server/diagrams/finishJob.puml
```diff
@@ -0,0 +1,43 @@
+@startuml
```
JohannesGaessler: This diagram is misguided.
Here is what should happen when a Job finishes:
* Server gets information about Job being finished
* Server re-evaluates which Jobs to run on which WorkerClients. This is the **exact same** process like when a Job is added by a user.

src/ja/server/Dispatcher.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def dispatch_job(job: Job) -> Message:
+        '''
+        Dispatches a job to a suitable work machine.
+
+        @param job Job to be dispatched.
+        @return Message containing the state of the job after it finishes.
+        '''
+
+    def check_resources(job: Job) -> bool:
```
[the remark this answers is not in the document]
ammen99: I'd say checking resources belongs to the scheduler (it should already have information about which tasks are on which machines, and how much resources each machine has).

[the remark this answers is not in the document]
JohannesGaessler: Now that I think about it we will probably need check_resources on both WorkerClient and Dispatcher.
WorkerClient.check_resources should return if the WorkerClient has enough resources to run the Job.
Dispatcher.check_resources should return if there is any WorkerClient with enough resources to run the Job.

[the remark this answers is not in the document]
ammen99: Why is the dispatcher involved at all? Also is WorkerClient the actual Worker client or is this just some class which represents the WC in the database?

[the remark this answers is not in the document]
JohannesGaessler: My bad, I meant to write Scheduler, not Dispatcher.
Also by WorkerClient I meant WorkerClient**Proxy**.

PR #6, opened 2019-12-07
add class stubs for classes involved in handling user commands

ammen99:
A few explanations about some of the classes here: https://github.com/DistributedTaskScheduling/documentation/pull/74

As we discussed, only class stubs are there, no methods (we cannot write methods because we do not have all the types defined yet).

### DIFF
--- a/src/ja/common/message.py
+++ b/src/ja/common/message.py
@@ -1,0 +1,21 @@
+'''
+The Message module provides the necessary interfaces that have to be
+implemented in order for a class to be transferable between the components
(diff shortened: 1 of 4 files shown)


src/ja/common/message.py
```diff
@@ -0,0 +1,21 @@
+'''
+The Message module provides the necessary interfaces that have to be
+implemented in order for an class to be transferable between the components
```
JohannesGaessler: an class -> a class

PR #8, opened 2019-12-08
Worker client draft

nikolatzotchev:
I just wanted to ask if I'm understanding correctly everything.
The WorkerProxy(not sure if it should be called that way) should be invoked from the ssh.
Worker just starts everything and creates a docker client (that the commands will use in there execute() method).

### DIFF
--- a/src/ja/worker/docker_events.py
+++ b/src/ja/worker/docker_events.py
@@ -1,0 +1,24 @@
+"""
+This module is responsible for the monitoring of the docker containers
+running on the worker machine.
+"""
+from typing import Dict
+
+
+class DockerEvents:
+    """
+    Monitors the containers on the current machine
+    """
+
+    def __init__(self) -> None:
+        """!
+        Starts the event loop
+        that will be monitoring the state of the containers.
+        """
+
+    def change_filters(self, filters: Dict[str, str]) -> None:
+        """!
+        Change on which events the server is going to be notified
+        @param filters: filter the events by event time, container or image
+            filters and events are described here https://docs.docker.com/engine/reference/commandline/events/
+        """
--- a/src/ja/worker/proxy/remote.py
+++ b/src/ja/worker/proxy/remote.py
@@ -1,0 +1,21 @@
+"""
+WorkerRemote is a single-class command line program.
+
+It queues the command on the worker socket.
+
+The commands are received on the standard input, and the worker response
+will be printed on the standard output.
+"""
+
+from ja.common.proxy.remote import Remote
+
+
+class WorkerRemote(Remote):
+    """
+    The main class of the program.
+    """
+
+    def __init__(self) -> None:
+        """!
+        When created the WorkerRemote starts listening to the stdin.
+        """
--- a/src/ja/worker/worker_client.py
+++ b/src/ja/worker/worker_client.py
@@ -1,0 +1,17 @@
+"""
+Contains classes needed by each WC for providing a DockerClient as well
+as notifying the CS
+"""
+
+
+class Worker:
+    """
+    The main class on the Worker Client
+    """
+    def __init__(self) -> None:
+        """!
+        Creates the DockerClient object used in this worker machine
+        The configuration file will be read.
+        The command handler should be created.
+        The DockerEvents should be started
+        """
--- a/src/ja/worker/proxy/message_handler.py
+++ b/src/ja/worker/proxy/message_handler.py
@@ -1,0 +1,19 @@
+"""
+This module contains classes and definitions related to handling
+WorkerCommands.
+"""
+from ja.common.proxy.command_handler import CommandHandler
+
+
+class WorkerCommandHandler(CommandHandler):
+    """
+    WorkerMessageHandler receives WorkerCommand and performs the corresponding
+    actions on the worker client.
+    The command handler creates a socket.
+    It starts listening on it for incoming commands.
+    """
+
+    def __init__(self) -> None:
+        """!
+        Creates the socket and listens on it.
+        """


src/ja/common/response_message.py
```diff
@@ -0,0 +1,11 @@
+"""
+This module provides the interface which all respond implement.
```
ammen99: all respond? Maybe you meants responses?

src/ja/common/response_message.py
```diff
@@ -0,0 +1,11 @@
...
+"""
+from ja.common.message import Message
+
+
+class RespondMessage(Message):
+    """
+    A base class for all response commands
+    A respond message is a message
```
ammen99: "A message is a message" anyway:)

src/ja/worker/worker_client.py
```diff
@@ -0,0 +1,64 @@
...
+class Worker:
+    """
+    The main class on the Worker Client
+    """
+
+    def __init__(self) -> None:
+        """!
+        Initializes the Worker, here the SSHConnection with the Server should
+        take place, initializing the DockerClient as well as starting the event loop in DockerEvents
+         that is going to be notifying the Server when necessary.
```
ammen99: Why the space in the beginning of the line?

src/ja/worker/worker_client.py
```diff
@@ -0,0 +1,64 @@
...
+        """
+
+    def send_respond_message(self, msg: RespondMessage) -> None:
+        """!
+        This method will take a ResponseMessage and will send it to the Central Server.
+        @param msg The @RespondMessage
+        """
+
+
+class DockerEvents:
```
ammen99: Actually I am not sure how to implement this class. I imagined we'd have some event loop and poll an fd or something?

nikolatzotchev: I tough we can use https://docker-py.readthedocs.io/en/stable/client.html#docker.client.DockerClient.events
and just filter for example when e container finishes.

ammen99: Didn't know the API looks like that.
Anyway I realized the model depends on what we decide (server asks for updates or wc notifies server)

src/ja/common/message/worker.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+    A base class for all worker commands. A worker command is an action which
+    is sent by the central server and performed on the worker client. WorkerCommands
+    typically execute some kind of docker commands on the worker client.
+    """
+
+
+class WorkerResponse(Response):
+    """
+    A base class for all worker client responses. A server response is a Response
```
JohannesGaessler: A server response -> A worker response

src/ja/worker/docker_events.py
```diff
@@ -0,0 +1,14 @@
+"""
+This module is responsible for the monitoring of the docker containers running on the worker machine
+"""
+
+
+class DockerEvents:
+    """
+    Monitors the containers no the current machine
```
JohannesGaessler: no -> on

src/ja/worker/proxy/remote.py
```diff
@@ -0,0 +1,14 @@
...
+LocalCentralServer is a single-class command line program.
+
+It queues the command on the worker socket.
+
+The commands are received on the standard input, and the worker response
+will be printed on the standard output.
+"""
+
+
+class LocalCentralServer:
```
JohannesGaessler: This should be called WorkerRemote and be a sub-class of the abstract Remote class in ja.common.proxy.remote

nikolatzotchev:
What do you mean by 

> You still need to specify the interfaces of these classes.

Do mean it like adding interfaces of the classes, or like adding more methods to the classes.

JohannesGaessler:
@nikolatzotchev I mean that you need to add method stubs.

nikolatzotchev:
But only public methods right, private methods will be made lather right?

JohannesGaessler:
Only public methods.

nikolatzotchev:
I think it will be best to give the execute command of the Worker commands the Worker instance as argument, that way the server can be able to control basically everything on the worker.
The DockerEvents class will use the [docker event](https://docs.docker.com/engine/reference/commandline/events/) command with a filter to get events only on specific criteria.
For python I think it will be best to [Docker SDK for Python ](https://docker-py.readthedocs.io/en/stable/index.html). I'm not so well informed about UNIX sockets (I'm reading some stuff right now) and I don't know if we will need a special class for them.

ammen99:
Is this a draft of is this ready for review&merge? (I see @JohannesGaessler already approved)

JohannesGaessler:
It seems that was a mistake on my part.
I was just going through the changes to see if they seem reasonable and I did not consider that there might still be things missing.

nikolatzotchev:
I totally forgot to remove the draft thing sorry.

src/ja/worker/docker_events.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def __init__(self) -> None:
+        """!
+        Starts the event loop
+        that will be monitoring the state of the containers.
+        """
+
+    def change_filters(self) -> None:
+        """!
+        Change on witch events the server is going to be notified
```
ammen99: witch -> which

ammen99: Also, how is the change done? Should there be a parameter describing what events to listen for?

src/ja/worker/docker_events.py
```diff
@@ -0,0 +1,21 @@
...
+    Monitors the containers on the current machine
+    """
+
+    def __init__(self) -> None:
+        """!
+        Starts the event loop
+        that will be monitoring the state of the containers.
+        """
+
+    def change_filters(self, filters: dict) -> None:
```
ammen99: There is no description of the filters parameter, I can't see how to use it.

src/ja/worker/docker_events.py
```diff
@@ -0,0 +1,23 @@
...
+        """!
+        Change on which events the server is going to be notified
+        :param filters: filter the events by event time, container or image
```
ammen99: we use `@` in front in the other places (see my and @JohannesGaessler 's pull requests)

PR #9, opened 2019-12-08
Classes for Proxies, Remotes, RemoteSockets, moved server classes

JohannesGaessler:
Created classes to go along with https://github.com/DistributedTaskScheduling/documentation/pull/76.
Subject to change as we have not yet reached an agreement regarding networking specifics.
Moved some of the server classes added by @ammen99 in this PR.

TODO: create UML class diagrams for these classes.


ammen99:
I know we agreed that things which are used by more than 1 component belong to `common/`, but I am no longer so sure about this. For example, the ServerCommand message is going to operate on the Database - so it is more like a server function. Anybody else having similar thoughts or is it just me that is confused?

ammen99:
Also, what is the point of the `from_user` sub-package? I think we can reuse the so-called `local_job_adder` for all the communications (the executable may need a better name though).

JohannesGaessler:
@ammen99 
>I know we agreed that things which are used by more than 1 component belong to common/, but I am no longer so sure about this.

First and foremost, packages are a way to organize the project.
We can sort our Python modules in any way that is convenient for us.

However, for an object oriented design packages should also show cohesion: classes in packages should mostly interact with each other, not with classes in other packages.
Unfortunately in the case of Messages this is not possible.
By definition these objects need to be used in more than one package to be useful.

Due to the way our software will be deployed three high-level packages arise naturally: user, server, worker.
For the communication between these packages we need Message objects.
We can put these Message objects either in one of the components that are communicating with each other or into an extra package (common).
No matter how we choose the cohesion will be low.

As such, I think that we should put Message objects into common because it will be the easiest way to handle dependencies.
**I think that we should not introduce dependencies of any kind between user, server, and worker.**
Such dependencies would make testing unnecessarily hard.
Instead the above components should depend only on common.
And the only way to do that is to put objects that are used by more than one component into common.

>Also, what is the point of the from_user sub-package?

The from_user package does not have a point.
I added this package before we discussed network communication in the meeting and forgot to remove it afterwards.

ammen99:
> As such, I think that we should put Message objects into common because it will be the easiest way to handle dependencies.
> **I think that we should not introduce dependencies of any kind between user, server, and worker.**
> Such dependencies would make testing unnecessarily hard.
> Instead the above components should depend only on common.
> And the only way to do that is to put objects that are used by more than one component into common.

This makes sense, thanks for the detailed explanation.

src/ja/server/proxy/message_handler.py
```diff
@@ -0,0 +1,13 @@
+"""
+This module contains classes and definitions related to handling
+ServerMessages.
+"""
+
+from src.ja.common.proxy.message_handler import MessageHandler
```
ammen99: Without the 'src.', just `ja.common....`

src/ja/common/message/server.py
```diff
@@ -0,0 +1,20 @@
+"""
+This module provides the interfaces for Messages passed between the user client
+and the central server.
+"""
+from .base import Command, Response
```
ammen99: As said in the other PR, I think we should use absolute paths everywhere.

PR #10, opened 2019-12-08
Config, UserClientCLIHandler class stubs

JohannesGaessler:
This PR adds class stubs for configs and the UserClientCLIHandler

### DIFF
--- a/src/ja/common/config.py
+++ b/src/ja/common/config.py
@@ -1,0 +1,13 @@
+from ja.common.message.base import Serializable
+
+
+class Config(Serializable):
+    """
+    Base class for hierarchical configuration (files). The properties of Config
+    objects are equivalent to the arguments of CLIs. Can override properties
+    of other Config objects of the same type (or a subtype). Provides the
+    option to only override unset properties.
+
+    Command line arguments override user-specified configuration files.
+    User-specified configuration files override global configuration files.
+    """
--- a/src/ja/user/cli.py
+++ b/src/ja/user/cli.py
@@ -1,0 +1,6 @@
+class UserClientCLIHandler:
+    """
+    Class for handling user input on the user client. Transforms command line
+    arguments into a UserConfig object. Loads user client config files from the
+    disk. Creates ServerCommand objects from the unified configs.
+    """
--- a/src/ja/server/config.py
+++ b/src/ja/server/config.py
@@ -1,0 +1,7 @@
+from ja.common.config import Config
+
+
+class ServerConfig(Config):
+    """
+    Config for the central server.
+    """
--- a/src/ja/user/config/__init__.py
+++ b/src/ja/user/config/__init__.py
@@ -1,0 +1,4 @@
+"""
+Package for user client Config classes. There is one base class and one
+sub-class for each command of the user client CLI.
+"""
--- a/src/ja/user/config/add.py
+++ b/src/ja/user/config/add.py
@@ -1,0 +1,7 @@
+from ja.user.config.base import UserConfig
+
+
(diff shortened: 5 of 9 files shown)


ammen99:
I thought configuration was the assignment of @jooe1 and @M1keReck

ammen99:
Also a general notice: You are including python packages with relative paths (like `..base`) shouldn't we include them like `ja.common.` ? You need to set your `PYTHONPATH`

JohannesGaessler:
@ammen99 

>I thought configuration was the assignment of @jooe1 and @M1keReck

It is their assignment to create method stubs. These are only class stubs.

>Also a general notice: You are including python packages with relative paths (like ..base) shouldn't we include them like ja.common. ? You need to set your PYTHONPATH

Does it matter whether imports are relative?

ammen99:
> Does it matter whether imports are relative?

This is mostly what style we will going to use. Also relative paths need you to track where your own file is, and if you use "asbolute" (not sure the proper term for python) then you clearly see where each class is coming from.

FWIW, internet recommends absolute imports as well: https://stackoverflow.com/questions/4209641/absolute-vs-explicit-relative-import-of-python-module

JohannesGaessler:
I made all imports absolute.

src/ja/common/config.py
```diff
@@ -0,0 +1,13 @@
+from ja.common.message import Serializable
+
+
+class Config(Serializable):
```
ammen99: I want to ask because obviously I don't understand what these classes are for.
I thought config classes are like default values for job, mount paths, etc. In the end, we'd transfer the Job itself, and not the config. So why is the config serializable at all? Is it just for reading it from a YAML file?

JohannesGaessler: Yes.

JohannesGaessler: Disregard my last comment.
We're going to transfer the Config which has a nested Job.
The Config will also contain metadata.

ammen99: I am not sure about this. What is the metadata?

src/ja/user/cli.py
```diff
@@ -0,0 +1,6 @@
+class UserClientCLIHandler:
+    """
+    Class for handling user input on the user client. Transforms raw user input
```
ammen99: Might make sense to clarify what "raw user input" is. When I hear it, I think about the "raw_input()" thing in python, but here you actually mean command line arguments, right?

src/ja/common/config.py
JohannesGaessler: I just realized my comment was again incorrect because I misused the word metadata.

We are going to transfer Message objects.
We are not going to transfer Config objects.

Config objects contain all the necessary information to encode a CLI command.
They can contain data like the verbosity level which does not have a direct influence on Command effects.
It would also be where we would put options that are client-side-only (like user credentials).

Message objects are Serializable because we are transferring them as YAML.
Config objects are Serializable because we need to store them as YAML files on disk.

Metadata is stored in the Message object.
As of right now we are not planning to add relevant amounts of metadata but this is where metadata would be added.

src/ja/user/cli.py
JohannesGaessler: Correct.
By "raw user input" I mean command line arguments.
It's "raw" input in the sense that it has not been processed yet and is therefore not directly usable.
Do you think "unprocessed user input" would be a better term?

ammen99: Why not just use "command line arguments"? Do we expect any other form of input?

ammen99: Or maybe command line input.

PR #18, opened 2019-12-14
Database

ammen99:
This is my work so far on the Database classes. It already contains my other PR #2 which I hope will be merged soon (I'll then rebase this PR on top of master).

The classes are more or less incomplete because I only wrote the "interface" which other components like the scheduler will use. However I do not think there is much more, since the only other potential classes are those who do the actual database connection and run queries. I hope to get some early feedback before I start them though.

Part of #15

### DIFF
--- a/src/ja/common/work_machine.py
+++ b/src/ja/common/work_machine.py
@@ -3,7 +3,7 @@ from ja.common.message.base import Serializable
 
 class ResourceAllocation(Serializable):
     """
-    Represents a group of on a work machine.
+    Represents a group of resources on a work machine.
     """
     def __init__(self, cpu_threads: int, memory: int, swap: int):
         """!
--- a/src/ja/common/workmachine.py
+++ b/src/ja/common/workmachine.py
@@ -1,0 +1,33 @@
+from ja.common.message.base import Serializable
+
+
+class ResourceAllocation(Serializable):
+    """
+    Represents a group of on a work machine.
+    """
+    def __init__(self, cpu_threads: int, memory: int, swap: int):
+        """!
+        Create the ResourceAllocation object.
+
+        @param cpu_threads The amount of CPU threads.
+        @param memory The amount of RAM, in MB.
+        @param swap The amount of swap space, in MB.
+        """
+
+    @property
+    def cpu_threads(self) -> int:
+        """!
+        @return The allocated CPU threads.
+        """
+
+    @property
+    def memory(self) -> int:
+        """!
+        @return The allocated RAM in MB.
+        """
+
+    @property
+    def swap(self) -> int:
+        """!
+        @return The allocated swap space in MB.
+        """
--- a/src/ja/server/database/database.py
+++ b/src/ja/server/database/database.py
@@ -33,7 +33,7 @@ class ServerDatabase(ABC):
         """
 
     @abstractmethod
-    def assing_job_machine(self, job: Job, machine: WorkMachine) -> None:
+    def assign_job_machine(self, job: Job, machine: WorkMachine) -> None:
         """!
         Update the job's assigned work machine in the database.
         This will cause a RuntimeError if the job is not in the database.
--- a/src/ja/server/database/database.py
+++ b/src/ja/server/database/database.py
@@ -8,8 +8,7 @@ from ja.server.database.types.work_machine import WorkMachine
 
 class ServerDatabase(ABC):
     """!
-    ServerDatabase is the interface which a database used by JobAdder needs to
-    implement.
+    ServerDatabase is the interface which a database used by JobAdder needs to implement.
     """
 
     @abstractmethod
@@ -18,16 +17,14 @@ class ServerDatabase(ABC):
         Load a job from the database by its id.
 
         @param job_id The id of the job.
-        @return The stored data about the job, or None if no such job was
-          found.
+        @return The stored data about the job, or None if no such job was found.
         """
 
     @abstractmethod
     def update_job(self, job: Job) -> None:
         """!
         Update the job in the database.
-        If no job with the id of @job has been stored up to now, a new entry
-        will be created in the database.
+        If no job with the id of @job has been stored up to now, a new entry will be created in the database.
 
         @param job The job to update stored data for.
         """
@@ -46,8 +43,7 @@ class ServerDatabase(ABC):
     def update_work_machine(self, machine: WorkMachine) -> None:
         """!
         Update the stored data about the given work machine.
-        If there is no entry in the database for the given work machine, a new
-        entry will be added.
+        If there is no entry in the database for the given work machine, a new entry will be added.
 
         @param machine The machine to update data for.
         """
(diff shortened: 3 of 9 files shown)


src/ja/common/workmachine.py
```diff
@@ -0,0 +1,33 @@
+from ja.common.message.base import Serializable
+
+
+class ResourceAllocation(Serializable):
+    """
+    Represents a group of on a work machine.
```
nikolatzotchev: of what :)

ammen99: resources :P

src/ja/server/database/database.py
```diff
@@ -0,0 +1,87 @@
...
+    def update_job(self, job: Job) -> None:
+        """!
+        Update the job in the database.
+        If no job with the id of @job has been stored up to now, a new entry
+        will be created in the database.
+
+        @param job The job to update stored data for.
+        """
+
+    def assing_job_machine(self, job: Job, machine: WorkMachine) -> None:
```
M1keReck: assing -> assigning :D

ammen99:
Some thoughts: to achieve low coupling, I think we should use the following design:

1. Commands update database
2. When database is updated it calls a callback which I called reschedule_callback. The idea is, we
  don't want the database to actually depend on the scheduler.
3. The scheduler gets called, and becomes a reference to the database that called it. It fetches information from the database, does computation, updates the database.
4. Scheduler also has a callback for dispatching, it then calls this callback.
5. Dispatcher does its thing.

In this design, the scheduler needs to know the information of the database. There are no other "API" dependencies, so I think we can't make it much better.

nikolatzotchev:
> The scheduler gets called, and becomes a reference to the database that called it. It fetches information from the database, does computation, updates the database.

But we have one database, when you speak about a database I image you are talking about maybe the ServerDatabase class or?

ammen99:
> But we have one database, when you speak about a database I image you are talking about maybe the ServerDatabase class or?

Sorry, I confused english and german :) The scheduler **gets** a reference to the database, I don't think it should actually store any information about it.

src/ja/server/database/types/work_machine.py
```diff
@@ -0,0 +1,95 @@
...
+        Initialize the RunningWorkMachineResources object. In the beginning,
+        all resources are marked as free.
+
+        @param total_resources The total amount of available resources.
+        """
+
+    @property
+    def total_resources(self) -> ResourceAllocation:
+        """!
+        @return The maximum resources available on the work machine.
```
nikolatzotchev: What exactly do you mean by maximum resources available? It is not very clear.

src/ja/server/database/types/work_machine.py
```diff
@@ -0,0 +1,95 @@
...
+        These resources will not be marked as free afterwards.
+
+        @param allocation The CPU, memory and swap space to allocate.
+        @param test_only If set to True, only the availability of enough
+          resources will be checked, but the resources will not be actually
+          allocated.
+        @return True if the allocation or the test was successful, False
+          otherwise.
+        """
+
```
nikolatzotchev: There should be a method that marks the resources as free, for example after a Job finishes we should mark the resources as free in the database.

src/ja/server/database/database.py
```diff
@@ -0,0 +1,99 @@
+from abc import ABC, abstractmethod
+from datetime import datetime
+from typing import List, Callable
+from ja.common.job import Job
+from ja.server.database.types.job_entry import DatabaseJobEntry
+from ja.server.database.types.work_machine import WorkMachine
+
+
+class ServerDatabase(ABC):
```
nikolatzotchev: There can also be a method for querying all jobs on a specific working machine.

src/ja/common/work_machine.py
```diff
@@ -0,0 +1,33 @@
+from ja.common.message.base import Serializable
+
+
+class ResourceAllocation(Serializable):
```
nikolatzotchev: I'm not so sure about the name, but maybe something like MachineResources is better. What do you think?

src/ja/server/database/database.py
ammen99: I'll add this as a parameter to query_jobs.

nikolatzotchev: Also probably a method for getting all workmachines with enough available resources(to run a job), the RunningWorkMachineResources may be passed as an argument.

src/ja/server/database/types/work_machine.py
```diff
@@ -0,0 +1,95 @@
...
+    A collection of the data needed to communicate with a work machine.
+    """
+    # TODO: this class is incomplete
+
+
+class WorkMachine:
+    """
+    Represents the data stored about a work machine in the database.
+    Each work machine is uniquely determined by its IP address.
+    (XXX: is the above ok?)
```
nikolatzotchev: What do we get by using the IP address as an Id, we can just have an incrementing integer for the ids.

src/ja/server/database/types/work_machine.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class WorkMachineConnectionDetails:
+    """
+    A collection of the data needed to communicate with a work machine.
+    """
+    # TODO: this class is incomplete
+
+
+class WorkMachine:
```
nikolatzotchev: Here you can add maybe also a method that returns how long a worker machine has been online. That can be maybe used for some kind of statistics.

src/ja/server/database/database.py
ammen99: > Also probably a method for getting all workmachines with enough available resources(to run a job), the RunningWorkMachineResources may be passed as an argument.

This is something the scheduler will do internally.

src/ja/common/work_machine.py
ammen99: I named it ResourceAllocation because this can also be just a group of resources allocated for a job, not only the resources on a machine.

src/ja/server/database/types/work_machine.py
ammen99: Maybe this can go into the connection details? E.g when the connection was established.

src/ja/server/database/types/work_machine.py
```diff
@@ -0,0 +1,115 @@
+from ja.common.work_machine import ResourceAllocation
+
+
+class RunningWorkMachineResources:
```
JohannesGaessler: Why did you call this class RunningWorkMachineResources and not just something like WorkerResources?

src/ja/server/database/types/work_machine.py
```diff
@@ -0,0 +1,115 @@
...
+        """!
+        Construct a new work machine object.
+        If the work machine is not online, then only statistics about it are
+        stored. Otherwise, the work machine also tracks available resources
+        and connection details.
+
+        @param uid The unique ID of the machine.
+        @param is_online Whether the work machine is online.
+        @param resources The available resources on the work machine.
+        @param connection The connection informatino about the work machine.
```
JohannesGaessler: informatino -> information

src/ja/server/database/database.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @abstractmethod
+    def query_jobs(self, since: datetime,
+                   user_id: int,
+                   work_machine: WorkMachine) -> List[DatabaseJobEntry]:
+        """!
+        Generate a list of all jobs belonging to @user_id since @since which
+        are or have been running on @workmachine.
+
+        @param since Date and time of the oldest interesting job.
```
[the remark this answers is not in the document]
ammen99: Makes sense to be able to do that.

src/ja/server/database/database.py
```diff
@@ -0,0 +1,103 @@
...
+        If there is no entry in the database for the given work machine, a new
+        entry will be added.
+
+        @param machine The machine to update data for.
+        """
+
+    @abstractmethod
+    def get_work_machines(self) -> List[WorkMachine]:
+        """!
+        @return A list of online work machines.
```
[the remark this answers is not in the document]
ammen99: online = connected right now, not at some point in the past

src/ja/server/database/database.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+
+    RescheduleCallback = Callable[['ServerDatabase'], None]
+
+    @abstractmethod
+    def set_scheduler_callback(self,
+                               callback: RescheduleCallback) -> None:
+        """!
+        Set a function which will be called whenever the database is updated
+        and this update makes rescheduling necessary.
```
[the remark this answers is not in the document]
ammen99: Hm, a good question. As it stands currently, we only need to connect one callback for the scheduler. However, we can implement the whole Observer pattern, in which case multiple callbacks can be connected. Would we have a need for anything like that?

src/ja/server/database/types/work_machine.py
ammen99: Renaming it to WorkMachineResources (I don't think we use Worker much)

src/ja/server/database/database.py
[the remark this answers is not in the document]
JohannesGaessler: Probably not as of right now, but I think our design will be judged based on how easy it can be expanded upon.
I think it's fine either way though.

PR #19, opened 2019-12-14
Method stubs for proxy-related classes

JohannesGaessler:
This PR primarily adds method stubs for objects related to communication between machines: Message, Command, Response, Remote, CommandHandler, SSHConnection, Proxy.


nikolatzotchev:
Can you just tell me why for example here

>      def from_dict(cls, property_dict: Dict[str, object]) -> "Serializable":

the Serializable is surrounded by "

JohannesGaessler:
@nikolatzotchev 
The quotation marks allow you to specify a type by name instead of by reference.
I need to do it in this case because I want to use the Serializable type while I'm defining the Serializable class.
At this point the Serializable type is not yet completely defined.
This causes a NameError when you try to actually define the classes (by running the module).

JohannesGaessler:
I have a similar problem when I want to use the type of a class that is defined further down:
Because the class is not yet defined I get a NameError.

src/ja/common/message/base.py
```diff
@@ -3,33 +3,97 @@
...
+    @abstractmethod
+    def from_dict(cls, property_dict: Dict[str, object]) -> "Serializable":
+        """!
+        @param property_dict A Python dictionary defining the desired
+        properties of the Serializable object to be created.
+        @return A new Serializable object based on the property entries of the
+        specified dictionary.
+        """
+
+    @abstractmethod
```
ammen99: Why is this and the `from_string` an abstract method? If we have conversion to dictionary, we can just use this conversion to stringify and convert back again.

src/ja/common/message/base.py
```diff
@@ -3,33 +3,97 @@
...
     dictionary. The dictionary should contain only basic python types:
-    strings, ints, bytes, doubles, lists and dictionaries.
+    strings, ints, bytes, doubles, lists and dictionaries. The dictionary can
+    also be dumped as/read from a YAML string.
     """
 
+    @abstractmethod
+    def to_dict(self) -> Dict[str, object]:
+        """!
+        @return A dictionary representation of this object.The entries in the
```
ammen99: I'd suppose YAML doesn't support all kind of objects as values. Maybe this is a good place to describe what kind of objects are allowed? OTOH we'd need ints, bools, strings and lists/dictionaries with those types (and dictionaries of dictionaries of those types, etc.)

src/ja/common/message/base.py
```diff
@@ -3,33 +3,97 @@
...
 
-class Command(Message):
+class Command(Message, ABC):
     """
     A base class for messages which are sent with the intent of performing an
     action on the remote component.
     """
 
+    @abstractmethod
+    def get_success_response(self) -> "Response":
```
ammen99: Can you elaborate what we'll use those `get_success_response` for?

src/ja/common/message/base.py
```diff
@@ -3,33 +3,97 @@
...
-class Response(Message):
+    @abstractmethod
+    def get_failure_response(self) -> "Response":
+        """!
+        @return a Response object indicating that the Command was not
+        successfully executed.
+        """
+
+
+class Response(Message, ABC):
```
ammen99: Is the double inheritence from ABC so that we can have the appropriate symbol in the class diagram?

src/ja/common/proxy/remote.py
```diff
@@ -6,9 +6,14 @@ class Remote(ABC):
...
-    on the Message object to a MessageHandler object via a socket. Remotes then
-    listen for a response (YAML string) on a socket. The response is then
-    written to stdout.
+    Receives a Command object represented by a YAML string on stdin. Asserts
+    that the user specified in the YAML document is the user calling the
+    Remote. Passes on the YAML string to a CommandHandler object via a socket.
+    Listens for a Response (as YAML string) on a different socket. Writes the
+    Response YAML string to stdout.
     """
+    def __init__(self, output_port: int, input_port: int):
```
ammen99: Again, ports or sockets? Also, to connect to a string, you need a string (when you create a socket, you get an FD, which you then bind to a file on the filesystem. When connecting to this socket, you need to provide the name, because you can't have an FD)

src/ja/worker/proxy/proxy.py
```diff
@@ -0,0 +1,40 @@
...
+    Proxy for the central server used on the worker client.
+    """
+    def __init__(self, ssh_config: SSHConfig):
+        """!
+        @param ssh_config: Config for paramiko
+        """
+
+    def register_self(self) -> ServerResponse:
+        """! Adds the worker client to the worker client pool of the central
+        server. The server will establish an SSHConnection to the worker client
```
ammen99: establish a connection to the server, not to the wc

src/ja/worker/proxy/proxy.py
```diff
@@ -0,0 +1,40 @@
...
+        and begin sending WorkerCommands.
```
ammen99: send or receive? I am confused.

src/ja/common/message/base.py
JohannesGaessler: It is literally impossible to define a finite type that does what we need.
The problem is that MyPy does not support recursive typing as of right now.
I'll explain on Monday.

src/ja/worker/proxy/proxy.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    Proxy for the central server used on the worker client.
+    """
+    def __init__(self, ssh_config: SSHConfig):
+        """!
+        @param ssh_config: Config for paramiko
+        """
+
+    def register_self(self) -> ServerResponse:
+        """! Adds the worker client to the worker client pool of the central
```
ammen99: Sometimes you start the documentation on a new line, sometimes not. Why?

src/ja/common/message/base.py
ammen99: I guessed as much, but I meant more like describing what we expect from the dictionary in the comments.

src/ja/common/message/base.py
JohannesGaessler: It's so that Response is an abstract class.
If you inherit from an abstract class the sub-class is not automatically an abstract class, after all.

src/ja/common/message/base.py
JohannesGaessler: Sorry, I misread your comment.
You're right, we should define what is allowed.

src/ja/common/proxy/command_handler.py
```diff
@@ -0,0 +1,25 @@
...
+    YAML string.
+    Afterwards the Command is validated semantically: the CommandHandler checks
+    if the user has the necessary permissions to perform the Command, if any
+    specified objects actually exist, etc.
+    If both validations pass, the Command is executed.
+    Always prints a Response as YAML string to the output_port at the end. The
+    success property of the Response is True if both validations were passed
+    and the Command was executed without error, and is false otherwise.
+    """
+    def __init__(self, input_port: int, output_port: int):
```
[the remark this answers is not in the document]
JohannesGaessler: Sockets need ports to operate.
We need one port for Remote -> CommandHandler and one for CommandHandler -> Remote.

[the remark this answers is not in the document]
ammen99: I guess we are speaking about different sockets. I am speaking about a unix named socket.

src/ja/common/proxy/remote.py
```diff
@@ -6,9 +6,14 @@ class Remote(ABC):
...
 
-    Remotes receive a Message object represented by a YAML string on stdin.
-    Remotes then convert the YAML string to a Message object. Remotes then pass
-    on the Message object to a MessageHandler object via a socket. Remotes then
-    listen for a response (YAML string) on a socket. The response is then
-    written to stdout.
+    Receives a Command object represented by a YAML string on stdin. Asserts
+    that the user specified in the YAML document is the user calling the
+    Remote. Passes on the YAML string to a CommandHandler object via a socket.
+    Listens for a Response (as YAML string) on a different socket. Writes the
```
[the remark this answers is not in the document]
JohannesGaessler: I was misinformed about sockets.
I thought they only allowed one-way communication.

src/ja/worker/proxy/proxy.py
JohannesGaessler: To send this command, the worked needs to have already set up the SSHConnection to the server.
This is so that the server then also creates an SSHConnection to the worker.

src/ja/worker/proxy/proxy.py
JohannesGaessler: No particular reason, I just didn't pay attention.

src/ja/worker/proxy/proxy.py
ammen99: I misread the comment, sorry.

src/ja/common/proxy/remote.py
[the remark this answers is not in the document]
ammen99: You can have bidirectional communication. You can also habe multiple clients on the same socket, and the daemon can differentiate between them and send each of them separate responses.

src/ja/common/message/base.py
JohannesGaessler: >I guessed as much, but I meant more like describing what we expect from the dictionary in the comments.

Didn't I already do that in the docstring of the Serializable class?

src/ja/common/message/base.py
JohannesGaessler: You're right.

src/ja/common/message/base.py
JohannesGaessler: Those methods are for creating the corresponding Response objects from a Command object.
The contents of a Response will vary depending on the Command.
The idea is that sub-classes will define how Responses will be constructed.

src/ja/common/message/base.py
```diff
@@ -3,33 +3,97 @@
...
+    @abstractmethod
+    def result_string(self) -> str:
+        """!
+        @return A human-readable string that states the result of the sent
+        Command.
+        """
+
+    @property
+    @abstractmethod
+    def success(self) -> bool:
```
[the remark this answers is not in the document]
JohannesGaessler: Fine by me.

src/ja/common/message/base.py
JohannesGaessler: Actually, now that I think about it, these objects should be private and therefore not part of the interface.
I'll remove them.

src/ja/common/message/server.py
```diff
@@ -1,20 +1,19 @@
...
 
-class ServerCommand(Command):
+class ServerCommand(Command, ABC):
     """
-    A base class for all server commands. A server command is an action which
-    is sent by the user client and performed on the server. ServerCommands
-    typically operate on the server database.
+    Base class for all server commands. A server command is an action which is
+    sent by a user client or the worker client and performed on the server.
+    ServerCommands typically operate on the server database.
```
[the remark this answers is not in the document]
JohannesGaessler: I'm adding execute methods for Commands.
I will add arguments like the database later on once we've merged everything.

src/ja/common/message/base.py
ammen99: oh yes, you did.

src/ja/common/message/base.py
```diff
@@ -3,33 +3,90 @@
...
 
-class Command(Message):
+class Command(Message, ABC):
     """
     A base class for messages which are sent with the intent of performing an
     action on the remote component.
     """
 
+    @abstractmethod
+    def execute(self, target: object) -> "Response":
```
ammen99: I do not think we can have such an execute() method without violating Liskov Substitution Principle, because subclasses need stronger preconditions on the `target` parameter.

src/ja/common/proxy/ssh.py
```diff
@@ -1,6 +1,99 @@
...
+
 class SSHConnection:
     """
-    Establishes an SSH connection to a remote component. Writes Command objects
-    the stdin of Remote objects. Then reads a Response object from stdout of
-    the Remote object. Message objects are read/written as YAML strings.
+    Establishes an SSH connection to a Remote object. Writes Command objects to
+    the stdin of said Remote objects. Then reads a Response object from stdout
+    of the Remote objects. Message objects are read/written as YAML strings.
+    Uses paramiko as the backend for establishing an ssh connection.
```
ammen99: Is the usage of paramiko really relevant to anybody reading this comment? We can mention it in the design document in a page "external libraries" or something like that.

src/ja/common/proxy/ssh.py
```diff
@@ -1,6 +1,99 @@
...
+
+    def send(self, command: Command) -> Response:
+        """!
+        Sends a Command to the Remote on the host. Automatically attaches
+        the username specified in ssh_config to the Command.
+        @param command: The Command to sent to the Remote.
+        @return: The Response received from the Remote.
+        """
+
+    def close(self, ) -> None:
```
ammen99: missing argument or why is there a comma here?

src/ja/server/proxy/proxy.py
```diff
@@ -0,0 +1,38 @@
...
+class WorkerProxy(ContinuousProxy):
+    """
+    Proxy for the worker client used on the central server.
+    """
+    def __init__(self, ssh_config: SSHConfig):
+        """!
+        @param ssh_config: Config for paramiko.
+        """
+
+    # TODO replace object with Job class
```
ammen99: Job was merged.

src/ja/worker/proxy/proxy.py
```diff
@@ -0,0 +1,42 @@
...
+class WorkerServerProxy(ContinuousProxy):
+    """
+    Proxy for the central server used on the worker client.
+    """
+    def __init__(self, ssh_config: SSHConfig):
+        """!
+        @param ssh_config: Config for paramiko
+        """
+
+    def register_self(self) -> ServerResponse:
```
ammen99: This should also have a parameter indicating the resources of the work machine.

ammen99:
By the way, why do we need the Remote object to be abstract (it was done in an earlier PR, but this one  changes it as well)? After all, all the remote does is get the user from the command, verify it and pass it further on a socket, this is the same on both the server and the worker.

src/ja/common/message/base.py
```diff
@@ -3,33 +3,90 @@
...
+
+class Message(Serializable, ABC):
     """
     A base class for all messages which can be transferred between the
     components of JobAdder.
     """
 
 
-class Command(Message):
+class Command(Message, ABC):
```
ammen99: Command should also have a user id, which the Remote uses for verification.

src/ja/common/message/base.py
```diff
@@ -3,33 +3,90 @@
...
+    @abstractmethod
+    def from_dict(cls, property_dict: Dict[str, object]) -> "Serializable":
+        """!
+        @param property_dict A Python dictionary defining the desired
+        properties of the Serializable object to be created.
+        @return A new Serializable object based on the property entries of the
+        specified dictionary.
+        """
+
+    def to_string(self) -> str:
```
M1keReck: It's convention to name this method __str__  in order to stick to the [Python data model](https://docs.python.org/3/reference/datamodel.html)

src/ja/common/message/base.py
JohannesGaessler: I think you're right.
How should we fix this?
Remove the Command super-class and inherit directly from Message?

src/ja/common/proxy/ssh.py
JohannesGaessler: It is relevant to our design because this makes it clear that we do not have to describe lower-level classes.

src/ja/common/proxy/ssh.py
JohannesGaessler: I'm not sure.
I think it was added automatically by my IDE.

src/ja/server/proxy/proxy.py
JohannesGaessler: This PR also depends on my class stubs UserClientCLI.
I'll rebase and fix this once that PR is merged.

src/ja/worker/proxy/proxy.py
JohannesGaessler: Added as TODO for now, I'll fix it when I rebase.

src/ja/common/message/base.py
JohannesGaessler: Should user id be of type str or int?

src/ja/common/message/base.py
JohannesGaessler: Good point.

src/ja/common/message/base.py
ammen99: Both are fine, but since the job's user id is an int, maybe int will be better for consistency

src/ja/common/message/base.py
ammen99: We should remove execute() from Comand, and have only the user id property (so se still have some function in the base class)

src/ja/common/message/base.py
nikolatzotchev: There is a to_string() function in ssh.py, that should also be renamed.

src/ja/common/proxy/remote.py
```diff
@@ -1,9 +1,6 @@
-from abc import ABC
-
-
-class Remote(ABC):
+class Remote(object):
```
ammen99: isn't deriving from object automatic in python?

nikolatzotchev: It does.

JohannesGaessler: object is indeed the default super-class.
I only added it for clarity.

src/ja/common/message/base.py
JohannesGaessler: I made it a string because for an SSH connection you are specifying a username, not an ID.

PR #20, opened 2019-12-15
Scheduler

### LINKED ISSUE #15: Scheduler and database class diagrams

ammen99:
This PR adds description of the Scheduler class.
It assumes that #18 is already merged, because it uses types defined in the Database PR.

Also I have put almost all of the logic in the Scheduler. It is responsible for freeing resources when a job is killed/done/etc. The database will simply store/update data, and also keep track of the runtime of each job.

Together with #18, fixes #15

### DIFF
--- a/src/ja/server/scheduler/algorithm.py
+++ b/src/ja/server/scheduler/algorithm.py
@@ -15,7 +15,7 @@ class CostFunction(ABC):
         """!
         Calculate the cost of a job.
 
-        The cost of a job is an integer. Higher values mean higher priority for
+        The cost of a job is a float. Lower values mean higher priority for
         the job when considering which jobs to schedule.
 
         @param job The job whose cost should be calculated.
@@ -26,7 +26,7 @@ class CostFunction(ABC):
     @property
     def blocking_threshold(self) -> float:
         """!
-        @return The minimum cost a job can have so that it can block further
+        @return The maximum cost a job can have so that it can block further
           jobs with smaller cost from being scheduled.
         """
 
@@ -34,8 +34,8 @@ class CostFunction(ABC):
     @property
     def preempting_threshold(self) -> float:
         """!
-        @return The minimum cost a job can have so that it can preempt already
-          running jobs. Should be at least @blocking_threshold.
+        @return The maximum cost a job can have so that it can preempt already
+          running jobs. Should be at most @blocking_threshold.
         """
 
 
--- a/src/ja/server/scheduler/algorithm.py
+++ b/src/ja/server/scheduler/algorithm.py
@@ -7,12 +7,11 @@ from ja.server.database.types.work_machine import WorkMachine
 
 class CostFunction(ABC):
     """
-    A CostFunction is a collection of functions which determine the order in
-    which jobs are scheduled.
+    A CostFunction is a function which determines the order in which jobs are scheduled.
     """
 
     @abstractmethod
-    def calculate_cost(self, job: Job) -> int:
+    def calculate_cost(self, job: Job) -> float:
         """!
         Calculate the cost of a job.
 
@@ -25,7 +24,7 @@ class CostFunction(ABC):
 
     @abstractmethod
     @property
-    def blocking_threshold(self) -> int:
+    def blocking_threshold(self) -> float:
         """!
         @return The minimum cost a job can have so that it can block further
           jobs with smaller cost from being scheduled.
@@ -33,7 +32,7 @@ class CostFunction(ABC):
 
     @abstractmethod
     @property
-    def preempting_threshold(self) -> int:
+    def preempting_threshold(self) -> float:
         """!
         @return The minimum cost a job can have so that it can preempt already
           running jobs. Should be at least @blocking_threshold.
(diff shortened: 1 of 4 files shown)


src/ja/server/scheduler/algorithm.py
```diff
@@ -0,0 +1,115 @@
+from abc import ABC, abstractmethod
+from typing import List, Optional, Tuple
+from ja.common.job import Job
+from ja.server.database.database import ServerDatabase
+from ja.server.database.types.work_machine import WorkMachine
+
+
+class CostFunction(ABC):
+    """
+    A CostFunction is a collection of functions which determine the order in
```
JohannesGaessler: Why a collection of functions?
Shouldn't it be just one function per cost function object and multiple cost function objects?

src/ja/server/scheduler/algorithm.py
```diff
@@ -0,0 +1,115 @@
...
+    which jobs are scheduled.
+    """
+
+    @abstractmethod
+    def calculate_cost(self, job: Job) -> int:
+        """!
+        Calculate the cost of a job.
+
+        The cost of a job is an integer. Higher values mean higher priority for
```
JohannesGaessler: Cost should be a float.

src/ja/server/scheduler/algorithm.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class CostFunction(ABC):
+    """
+    A CostFunction is a collection of functions which determine the order in
+    which jobs are scheduled.
+    """
+
+    @abstractmethod
+    def calculate_cost(self, job: Job) -> int:
```
JohannesGaessler: What about the parameters for the cost function?
Are you going to pass them to the CostFunction constructor?

src/ja/server/scheduler/algorithm.py
```diff
@@ -0,0 +1,115 @@
...
+        The cost of a job is an integer. Higher values mean higher priority for
+        the job when considering which jobs to schedule.
+
+        @param job The job whose cost should be calculated.
+        @return The calculated cost of the job.
+        """
+
+    @abstractmethod
+    @property
+    def blocking_threshold(self) -> int:
```
JohannesGaessler: Why is blocking threshold a property of the cost function?
Shouldn't it be a property of the Scheduler?

src/ja/server/scheduler/algorithm.py
```diff
@@ -0,0 +1,115 @@
...
+        """!
+        @return The minimum cost a job can have so that it can block further
+          jobs with smaller cost from being scheduled.
+        """
+
+    @abstractmethod
+    @property
+    def preempting_threshold(self) -> int:
```
JohannesGaessler: See blocking_threshold

src/ja/server/scheduler/algorithm.py
```diff
@@ -0,0 +1,115 @@
...
+        """!
+        @return The minimum cost a job can have so that it can preempt already
+          running jobs. Should be at least @blocking_threshold.
+        """
+
+
+class JobDistributionPolicy(ABC):
+    """
+    JobDistributionPolicy is a collection of functions which determine how
+    jobs to be scheduled are to be distributed among work machines.
```
JohannesGaessler: Wouldn't it make more sense to make one JobDistributionPolicy super class that has three sub-classes for regular assigning, blocking, and preemption instead of one class with three methods?

src/ja/server/scheduler/scheduler.py
```diff
@@ -0,0 +1,30 @@
...
+from ja.server.scheduler.algorithm import SchedulingAlgorithm
+
+
+class Scheduler:
+    """
+    Scheduler is a class which is responsible for generating the new job
+    distribution whenever the database changes.
+
+    Internally it uses a SchedulingAlgorithm to determine how to schedule
+    jobs which are "alive".
```
JohannesGaessler: "Alive" is not precise enough I think.

src/ja/server/scheduler/algorithm.py
ammen99: I initially had another design but later changed it, and forgot to update comments. It is as you say, a single cost function.

src/ja/server/scheduler/algorithm.py
ammen99: CostFunction is an abstract class. Any specific constants belong to the concrete implementation of that cost function.

src/ja/server/scheduler/algorithm.py
ammen99: I wanted to make CostFunction control how jobs are scheduled (ordering and blocking/preempting policy). Maybe CostFunction is not a good name at all and that's why it is unclear.

src/ja/server/scheduler/algorithm.py
ammen99: I did it this way because otherwise some of the returned values won't be used in all cases, but maybe it doesn't matter. I'll change it as you say.

src/ja/server/scheduler/scheduler.py
ammen99: is `runnable` better? This is what I have used below in `reschedule`

src/ja/server/scheduler/algorithm.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    A CostFunction is a function which determines the order in which jobs are scheduled.
+    """
+
+    @abstractmethod
+    def calculate_cost(self, job: Job) -> float:
+        """!
+        Calculate the cost of a job.
+
+        The cost of a job is an integer. Higher values mean higher priority for
```
JohannesGaessler: Docstring still says integer.
Also, going by the convention for cost functions, the Scheduler would be supposed to pick the job with the lowest cost function value, not the one with the highest.
(We can just use negative effective priority for this.)

src/ja/server/scheduler/algorithm.py
JohannesGaessler: I think I understand now why you've made the thresholds part of the cost function.
The thresholds must be relative to the values produced by the cost function.
And because the specific values produced by a cost function depend on the implementation of the cost function, the thresholds should be bundled together with the cost function, correct?
I agree that the thresholds should be placed in the cost function.

ammen99: Yes, you're correct.

src/ja/server/scheduler/algorithm.py
ammen99: Fixed.

src/ja/server/scheduler/default_policies.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def assign_machine(self,
+                       job: Job,
+                       distribution: ServerDatabase.JobDistribution,
+                       available_machines: List[WorkMachine]) -> Optional[Tuple[WorkMachine, List[Job]]]:
+        # Should iterate over all available machines and select the best one depending on _assign_machine_cost
+        pass
+
+
+class DefaultNonpreemptiveDistributionPolicy(DefaultJobDistributionPolicyBase):
```
nikolatzotchev: DefaultNonpreemptiveDistributionPolicy -> DefaultNonPreemptiveDistributionPolicy 
I'm not sure about this. It is spelled non-preemptive right, you already named the variables nonpreemtive, maybe they should be also changed to non_preemtive. I'm not sure.

ammen99: Fixed.

PR #21, opened 2019-12-15
TravisCI: ignore F821 in flake8

ammen99:
Why? If you look at https://github.com/DistributedTaskScheduling/JobAdder/pull/18, you will see that the build fails, because flake8 reports some types as not defined. However, they are defined - on the line above! I think this is a bug in flake8 because MyPy does not report any error and MyPy also does check for undefined types.

I will open an issue in flake8 and see how it goes, but in the meantime we can just disable the check.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,6 @@ python: 3.6
 
 install: "pip3 install flake8 mypy"
 script:
-  - flake8 src/
+  - flake8 src/ --ignore F821
   - ./.travis/check_init_py.sh src/
   - MYPYPATH=$(pwd)/src mypy -p ja --no-strict-optional --strict


ammen99:
Reported the bug here: https://github.com/PyCQA/pyflakes/issues/496

ammen99:
Turns out that the bug is already fixed, but the bug fix is not in any release yet so we still need it until a new pyflakes release comes out.

JohannesGaessler:
Have you tried running the Python file that you have defined those classes in?
I think it will throw a NameError regardless of what flake8 has to say because you're using a type before you have finished defining it (I fixed this by putting the type in quotation marks, but we will need to patch py_puml_tools for that).

ammen99:
> Have you tried running the Python file that you have defined those classes in?
> I think it will throw a NameError regardless of what flake8 has to say because you're using a type before you have finished defining it (I fixed this by putting the type in quotation marks, but we will need to patch py_puml_tools for that).

Hm, I haven't tried to run Python, but I assumed it would work because MyPy accepts the file ...

PR #24, opened 2020-01-07
travis.yml: fix max line length setting

ammen99:
I don't know what happened the first time, but I got it wrong. It should be `--max-line-length=120`, not `--config max-line-length 120`

Basically fixes #22
PS: this time, somebody please test that I haven't done something stupid again :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,6 @@ python: 3.6
 
 install: "pip3 install flake8 mypy"
 script:
-  - flake8 src/ --ignore F821 --config max-line-length 120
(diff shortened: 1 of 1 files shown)


ammen99:
> I checked the [flake8 documentation](https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-max-line-length) and this time I think you are right. I don't have an idea how to check it locally but I thing it should be fine.

You can just create a file with a line with length 100 (or anything between 80 and 120) and check that flake8 doesn't output an error.

PR #25, opened 2020-01-07
PlantUML image size limit

JohannesGaessler:
It seems we forgot to override the PlantUML image size limit of 2048x2048.
This PR fixes that.


JohannesGaessler:
This also works.

PR #28, opened 2020-01-09
Notifier web api

### LINKED ISSUE #23: Email Notifier and Web API
I think I was assigned these two in our last meeting.

Two questions:

1. Do we store a list of emails for each user, or do we store an email for each job?
2. Do we need HTTPS for the server or plain HTTP?

ammen99:
Fixes #23

### DIFF
--- a/src/ja/common/job.py
+++ b/src/ja/common/job.py
@@ -63,12 +63,14 @@ class Job:
     """!
     Represents a job with all of its attributes.
     @param owner_id The unix user id of the user who owns this job.
+    @param email The email of the user who created the job, or None.
     @param priority The priority of the job.
     @param ctx The docker context of the job.
     @param constraints The docker constraints of the job.
     """
     def __init__(self,
                  owner_id: int,
+                 email: str,
                  constraints: JobSchedulingConstraints,
                  ctx: IDockerContext,
                  docker_constraints: DockerConstraints):
@@ -94,6 +96,12 @@ class Job:
     def owner_id(self) -> int:
         """!
         @return The id of the user who owns the job.
+        """
+
+    @property
+    def email(self) -> str:
+        """!
+        @return The email of the user who owns the job, or None if no email was specified.
         """
 
     @property
--- a/src/ja/server/web/requests.py
+++ b/src/ja/server/web/requests.py
@@ -1,0 +1,111 @@
+from abc import ABC, abstractmethod
+from ja.common.message.base import Serializable
+from ja.server.database.database import ServerDatabase
+from typing import Dict
+
+
+class WebRequest(Serializable, ABC):
+    """
+    A base class for the different statistics requests.
+    """
+
+    @abstractmethod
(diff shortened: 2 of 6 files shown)


src/ja/common/job.py
```diff
@@ -63,12 +63,14 @@ class Job:
     """!
     Represents a job with all of its attributes.
     @param owner_id The unix user id of the user who owns this job.
```
nikolatzotchev: unix -> Unix

ammen99: This is not a part of this PR at all. Send another pull request to fix it :)

src/ja/server/web_api.py
```diff
@@ -0,0 +1,33 @@
...
+class StatisticsWebServer:
+    """
+    Creates a web server which enables external applications to obtain statistics about JobAdder.
+    """
+
+    def __init__(self, host: str, port: int, database: ServerDatabase):
+        """!
+        Initialize the web server.
+
+        @param host The host for the server.
```
JohannesGaessler: What do you mean by host?

ammen99: Sorry, I haven't properly checked the documentation, it is called server_name: https://docs.python.org/3/library/http.server.html#http.server.HTTPServer

ammen99:
I have imagined that we're going to have different classes for each request to the web server, just like we have different classes for commands, so I have added a class for each command. The list of requests is taken from https://github.com/DistributedTaskScheduling/documentation/pull/82

src/ja/server/web/requests.py
```diff
@@ -0,0 +1,111 @@
...
+class WorkMachineWorkloadRequest(WebRequest):
+    """
+    Generates the response to the machine workload request.
+    """
+
+    def __init__(self, uid: str):
+        """!
+        Initialize the request response.
+
+        @param uid The uid of the work machine this request is for.
```
JohannesGaessler: Can this be None to get the workload of all machines?

src/ja/server/web/requests.py
```diff
@@ -0,0 +1,111 @@
...
+class JobInformationRequest(WebRequest):
+    """
+    Generates the response to the job information request.
+    """
+
+    def __init__(self, uid: str):
+        """!
+        Initialize the request response.
+
+        @param uid The uid of the job the report is for.
```
JohannesGaessler: Can this be None to get a report for all jobs?

src/ja/server/web/requests.py
ammen99: I don't think so.

src/ja/server/web/requests.py
ammen99: Again, I don't think so.

src/ja/server/web/requests.py
ammen99: You can use the workload request to get the IDs of all machines, and then you can issue a request for each machine to get the running jobs, if that's what the user needs.

PR #31, opened 2020-01-09
created the dispatcher class

jooe1:


src/ja/server/dispatcher/Dispatcher.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    Represents the mediator between the Scheduler and the work machines.
+    """
+
+    def __init___(self):
+        """!
+        Constructor for the dispatcher class
+        """
+
+    def set_distribution(self, jobs: list[Job]) -> None:
```
ammen99: There is a special type for job distribution: `ServerDatabase.JobDistribution`
A list of jobs is not enough because a `Job` does not contain its assigned machine.

src/ja/server/dispatcher/Dispatcher.py
```diff
@@ -0,0 +1,36 @@
...
+        """!
+        Distributes the newly added jobs to the work machines
+        and potentially pauses jobs that need to be paused.
+
+        @param jobs list of the new job distribution.
+        """
+
+    def _get_changed_jobs(self) -> list[Job]:
```
ammen99: I suppose you'd need to pass the new distribution of jobs as a parameter so that this function can compare the old and the new distributions?

src/ja/server/dispatcher/Dispatcher.py
```diff
@@ -0,0 +1,36 @@
...
+        @return list of the jobs whose state has been changed.
+        """
+
+    def _reset(self) -> None:
+        """!
+        Dispatches the added jobs, pauses the paused jobs and cancels the cancelled jobs.
+        """
+
+    @property
+    def jobs(self):
```
ammen99: Why is this property needed? I think that the scheduler simply gives the dispatcher the next distribution, there's no need to get the information from the dispatcher.

src/ja/server/dispatcher/Dispatcher.py
```diff
@@ -0,0 +1,43 @@
...
+
+JobDistribution = ServerDatabase.JobDistribution
+
+
+class Dispatcher:
+    """
+    Represents the mediator between the Scheduler and the work machines.
+    """
+
+    def __init__(self, worker_proxies: List[WorkerProxy]):
```
ammen99: The list of proxies may change because new machines can be added and removed at any time.

Here's the solution that we thought of: There should be another class, perhaps with the name `WorkerProxyFactory`. It has one method which is `get_proxy` and takes a parameter the WorkMachine from the database. It returns a WorkerProxy for this WorkMachine.

This WorkerProxyFactory class also cleans up work machines which are no longer connected (this should be just mentioned in the class description, no public methods for this)

jooe1: would it be wrong if we use a setter for the list of proxies, or just give each instance of the work machine a proxy variable that we can access, or store the proxy details in the database and access it from there

ammen99: The connection details are already available in the WorkMachine (like IP address). However, we cannot store the Proxy object in the database (because it is a Python object, not some value).

We could have a setter for proxies, but I think it would be better if we can have an AbstractFactory which simply gives back the appropriate proxy.

If you are unsure about what I mean, I could send it in another pull request, it is a simple addition.

src/ja/server/dispatcher/Dispatcher.py
```diff
@@ -0,0 +1,43 @@
...
+        """!
+        Distributes the newly added jobs to the work machines.
+        and potentially pauses jobs that need to be paused.
+
+        @param jobs list of the new job distribution.
+        """
+
+    def _get_changed_jobs(self, jobs: List[Job]) -> List[Job]:
+        """!
+        @return list of the jobs whose state has been changed.
```
JohannesGaessler: Changed relative to what?
The last time the set_distribution was called?

jooe1: yes

src/ja/server/dispatcher/Dispatcher.py
jooe1: still not sure why we need to make another class that returns only 1 variable that we can just store in the work machine class

ammen99: The other class will also manage the internal cache of proxies, and also handle work machine disconnecting (it needs to update database in that case)

ammen99:
Something else I noticed, the file name should be 'dispatcher.py' not 'Dispatcher.py' for consistency with the rest of the project.

PR #33, opened 2020-01-09
Method stubs for config classes

JohannesGaessler:
This PR specifies the interfaces for config classes.

### DIFF
--- a/src/ja/common/job.py
+++ b/src/ja/common/job.py
@@ -1,9 +1,10 @@
 """
 This module contains the definition of the Job class and related structures.
 """
-from typing import List
+from typing import List, Dict
 from enum import Enum
 from ja.common.docker_context import DockerConstraints, IDockerContext
+from ja.common.message.base import Serializable
 
 
 class JobStatus(Enum):
@@ -29,14 +30,14 @@ class JobPriority(Enum):
     URGENT = 3
 
 
-class JobSchedulingConstraints:
+class JobSchedulingConstraints(Serializable):
     """
     JobConstraints is a collection of job properties which affect how it is
     scheduled.
     """
     def __init__(self,
                  priority: JobPriority,
-                 preemtible: bool,
+                 is_preemtible: bool,
                  special_resources: List[str]):
         pass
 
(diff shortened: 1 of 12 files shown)


src/ja/common/job.py
```diff
@@ -92,6 +99,11 @@ def uid(self, value: str) -> None:
         @param value The new job UID.
         """
 
+    def label(self) -> str:
```
nikolatzotchev: Shouldn't the label also be a property of the job?

nikolatzotchev: I mean why is there no label parameter in the __init__ function?

JohannesGaessler: Sorry, I just forgot to add it.

src/ja/server/config.py
```diff
@@ -1,7 +1,37 @@
+from typing import Dict
+
 from ja.common.config import Config
 
 
 class ServerConfig(Config):
```
ammen99: Several missing configuration options:

1. Amount of special resouces (number of licenses, etc)
2. Server port?
3. Database connection information? Or where should it come from?

JohannesGaessler: What exactly do you mean by database connection information?
Is it specified somewhere in the code?

ammen99: I am not very sure actually, because I am not very familiar witth databases.
But I assumed you do have some sort of username/password or at least a database identifier, right?

ammen99: I don't know if we're going to use PostreSQL, but just an example: https://wiki.postgresql.org/wiki/Using_psycopg2_with_PostgreSQL

There, you need a host, database name, user name and password to connect to the database.

ammen99: Or maybe we'll hardcode them?

nikolatzotchev: I think most of the times a configuration file. But we can also hardcode them.

JohannesGaessler: I'll just add another config.
It's not much work (both design and implementation).

JohannesGaessler: What are we going to need a server port for?

ammen99: Oops, didn't see this when merging. Basically, we need a port for the web api.

PR #37, opened 2020-01-12
Set PlantUML direction to left to right

JohannesGaessler:
Although this is not assigned to me, I looked up if it is possible to make PlantUML better with minimal effort for debugging.
As it turns out the formatting of our class diagram can be significantly improved by just setting `left to right direction`.
This PR adds this change.

main


ammen99:
I think we can really improve the layout of our class diagram in another way: making associations come before inheritance relations in the PUML file (I'll have to do a small patch to py2puml for that). I'll try that tomorrow.

ammen99:
I doubt we'll need this.

Issue #35, opened 2020-01-11
Make all proxies abstract base classes


ammen99:
We'll most probably need to have dummies here for the unit tests, so the WorkerServerProxy, etc. need to be interfaces.

PR #38, opened 2020-01-12
Dummy classes for proxy objects to be used for testing

JohannesGaessler:
Fixes https://github.com/DistributedTaskScheduling/JobAdder/issues/35


src/ja/server/proxy/proxy.py
```diff
@@ -1,38 +1,80 @@
...
+        pass
+
+    def pause_job(self, uid: str) -> WorkerResponse:
+        pass
+
+    def resume_job(self, uid: str) -> WorkerResponse:
+        pass
+
+
+class WorkerProxyDummy(IWorkerProxy):
```
ammen99: I think the dummies belong to the unit or integration tests, not to the source file?

JohannesGaessler:
Tests are in dedicated files.
You're probably right that the dummies should be put there.

PR #40, opened 2020-01-12
dispatcher: Use an abstract factory for generating the WorkerProxies

ammen99:


src/ja/server/dispatcher/proxy_factory.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    A concrete factory for creating and managing work proxies over the network.
+    Whenever a connection terminates, the database will be updated.
+    """
+
+    def __init__(self, database: ServerDatabase):
+        """!
+        Create a new WorkerProxyFactory.
+
+        @param database The database to update if a connection terminates.
```
JohannesGaessler: if -> when

PR #41, opened 2020-01-12
Sub-classes for ServerCommand

JohannesGaessler:
I noticed that we did not add classes for the messages sent to the server.
This PR adds those classes.

### DIFF
--- a/src/ja/server/database/types/work_machine.py
+++ b/src/ja/server/database/types/work_machine.py
@@ -1,3 +1,5 @@
+from enum import Enum
+
 from ja.common.work_machine import ResourceAllocation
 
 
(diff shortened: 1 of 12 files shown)


src/ja/server/database/types/work_machine.py
```diff
@@ -99,6 +99,14 @@ def online(self) -> bool:
           False otherwise.
         """
 
+    @property
```
ammen99: I would prefer to see the job machine as a state (ONLINE, RETIRED, OFFLINE) and then removing the online attribute.
Otherwise we have 2 boolean flags but not all of the 4 states are valid (i.e not retired but offline)

src/ja/server/database/types/work_machine.py
```diff
@@ -64,14 +66,20 @@ class WorkMachineConnectionDetails:
     # TODO: this class is incomplete
 
 
+class WorkMachineState(Enum):
+    ONLINE = 0
```
ammen99: not a problem, just curious: why 0, 10 and 20 instead of 0,1,2?

JohannesGaessler: I wasn't 100% if we will need more states.
I figured that if I leave some space in between states it would be easier to add more later.

PR #44, opened 2020-01-12
Add from_dict.

nikolatzotchev:
These all should implement from_dict right?


ammen99:
I do not like this. Perhaps we should add a class Deserializable and then Message should inherit from both?

Reason: `from_dict` makes zero sense for these requests.

JohannesGaessler:
Well, I was assuming the implementation for from_dict would be to just raise a NotImplementedError.
If you don't implement all methods you will get a RuntimeException on import.

nikolatzotchev:
Something like this should work, if we don't need everything from serializable we just don't use it.

PR #54, opened 2020-01-13
Implementation of Job class

JohannesGaessler:
This PR adds an implementation for the Job class and its sub-classes.
I also laid some groundwork to be used for serialization in general.

- [x] Implement Job.
- [x] Implement JobSchedulingConstraints.
- [x] Implement DockerContext
- [x] Implement DockerConstraints
- [x] Add unit tests.

### DIFF
--- a/src/ja/common/docker_context.py
+++ b/src/ja/common/docker_context.py
@@ -24,6 +24,12 @@ class MountPoint(Serializable):
         self._source_path = source_path
         self._mount_path = mount_path
 
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, MountPoint):
+            return self.source_path == other.source_path and self.mount_path == other.mount_path
+        else:
+            return False
+
     @property
     def source_path(self) -> str:
         """!
@@ -43,10 +49,10 @@ class MountPoint(Serializable):
 
     @classmethod
     def from_dict(cls, property_dict: Dict[str, object]) -> "MountPoint":
-        _source_path = cls._get_str_from_dict(property_dict=property_dict, key="source_path")
-        _mount_path = cls._get_str_from_dict(property_dict=property_dict, key="mount_path")
+        source_path = cls._get_str_from_dict(property_dict=property_dict, key="source_path")
+        mount_path = cls._get_str_from_dict(property_dict=property_dict, key="mount_path")
         cls._assert_all_properties_used(property_dict)
-        return MountPoint(source_path=_source_path, mount_path=_mount_path)
+        return MountPoint(source_path=source_path, mount_path=mount_path)
 
 
 class IDockerContext(Serializable, ABC):
@@ -54,6 +60,12 @@ class IDockerContext(Serializable, ABC):
     A docker context consists of the necessary data to build a docker image to
     run a job in.
     """
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, IDockerContext):
+            return self.dockerfile_source == other.dockerfile_source and self.mount_points == other.mount_points
+        else:
+            return False
 
     @property
     @abstractmethod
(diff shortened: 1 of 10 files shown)


src/ja/common/docker_context.py
```diff
@@ -21,18 +21,38 @@ def __init__(self, source_path: str, mount_path: str):
...
+        return self._source_path
 
     @property
     def mount_path(self) -> str:
         """!
         @return The path where the directory should be mounted.
         """
+        return self._mount_path
+
+    def to_dict(self) -> Dict[str, object]:
```
M1keReck: `return self.__dict__ `is enough for all the to_dict methods. Only difference is that the names of the attributes in the dictionary are going to start with an underscore.

JohannesGaessler: No it's not.
`self.__dict__` will directly return _all_ attributes set for an object, including attributes that are not relevant for the state of the object.
You would also still have to convert any non-primitive attribute into a dictionary so you could only use this implementation for the most simple of objects.

src/ja/common/job.py
```diff
@@ -39,58 +39,103 @@ def __init__(self,
...
                  owner_id: int,
                  email: str,
-                 constraints: JobSchedulingConstraints,
-                 ctx: IDockerContext,
+                 scheduling_constraints: JobSchedulingConstraints,
+                 docker_context: IDockerContext,
                  docker_constraints: DockerConstraints,
                  label: str = None):
-        pass
+        self._uid: str = None
```
nikolatzotchev: How will we generate the uid, i fount [this ](https://docs.python.org/3/library/uuid.html) online, but maybe we can just start with 1 and increment for each new job.

nikolatzotchev: > "If the job has not been queued yet, it has no UID." aa I see.

nikolatzotchev:
@JohannesGaessler I wanted to ask, what IDE are you using?

JohannesGaessler:
I am using PyCharm Community Edition.

nikolatzotchev:
Do you know how to run every test in a dir, I tried also in the terminal using 
> python -m unittest discover <test_directory>
but it runs 0 test, the same in pycharm.

JohannesGaessler:
>Do you know how to run every test in a dir, I tried also in the terminal using

I usually just do it via `setup.py`
I pushed a fixup where I quickly threw something together.
To run all tests, run `python setup.py test`

src/ja/common/job.py
```diff
@@ -146,22 +204,55 @@ def status(self, new_status: JobStatus) -> None:
...
+        _dict["docker_context"] = self.docker_context.to_dict()
+        _dict["docker_constraints"] = self.docker_constraints.to_dict()
+        _dict["status"] = self.status
+        _dict["label"] = self.label
+        return _dict
 
     @classmethod
     def from_dict(cls, property_dict: Dict[str, object]) -> "Job":
-        pass
+        _uid = cls._get_str_from_dict(property_dict=property_dict, key="uid", mandatory=False)
```
M1keReck: Shouldn't we specify the type of these local variables?
`_uid: str = cls._get_str_from_dict(property_dict=property_dict, key="uid", mandatory=False)`

JohannesGaessler: Because `_get_str_from_dict` always returns an object of type str the type of `_uid` is already known to be str.
For this reason type hints are optional here.
Do you think the code would be easier to understand if there were type hints?

M1keReck: No, it wouldn't improve readability, but I thought that typing is always mandatory in PSE.

> Because _get_str_from_dict always returns an object of type str the type of _uid is already known to be str.
In most strongly typed languages, you'd have to specify the type nevertheless.

Let's just wait for more opinions...

nikolatzotchev: I think it is fine, no one will care much about this, outside of us, so readability is more important.

src/test/serializable/base.py
```diff
@@ -0,0 +1,46 @@
...
+            _object_dict_copy = deepcopy(self._object_dict)
+            _object_dict_copy.pop(_key)
+            if _key in self._optional_properties:
+                self._object.__class__.from_dict(_object_dict_copy)
+            else:
+                with self.assertRaises(ValueError):
+                    self._object.__class__.from_dict(_object_dict_copy)
+
+    def test_adding_property_raises_error(self) -> None:
+        self._object_dict["AYY"] = "LAMO"
```
ammen99: `LMAO`  lol

nikolatzotchev: lol

src/ja/common/docker_context.py
```diff
@@ -41,19 +61,83 @@ class IDockerContext(Serializable, ABC):
...
+        return self._mount_points
+
+    def to_dict(self) -> Dict[str, object]:
+        _dict: Dict[str, object] = dict()
+        _dict["dockerfile_source"] = self.dockerfile_source
+        _dict["mount_points"] = [_mount_point.to_dict() for _mount_point in self.mount_points]
+        return _dict
+
+    @classmethod
+    def from_dict(cls, property_dict: Dict[str, object]) -> "IDockerContext":
```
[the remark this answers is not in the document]
JohannesGaessler: Doesn't need to be in quotes.
That is just the default type PyCharm gives you when you auto-add stubs for abstract methods and I forgot to change it.

src/ja/common/message/base.py
```diff
@@ -15,6 +15,71 @@ class Serializable(ABC):
...
+        if mandatory and not isinstance(_property, list):
+            cls._raise_error_wrong_type(key=key, expected_type="List[str]", actual_type=_property.__class__.__name__)
+        _list = cast(List[object], _property)
+        for _element in _list:
+            if not isinstance(_element, str):
+                cls._raise_error_wrong_type(key=key, expected_type="List[str]", actual_type="List[object]")
+        return cast(List[str], _list)
+
+    @classmethod
+    def _get_int_from_dict(cls, property_dict: Dict[str, object], key: str, mandatory: bool = True) -> int:
```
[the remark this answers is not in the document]
JohannesGaessler: I tried an implementation using generics, I couldn't get it to work.
The problem still is that typing.cast does not accept variable types at all.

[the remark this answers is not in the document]
ammen99: wait, why is cast needed here at all?

src/test/serializable/base.py
JohannesGaessler: LAMO is the ascended version of the meme.

ammen99: would like to see a reference to this ascended version :P

ammen99: or maybe we make it ascend, forget about it

JohannesGaessler: Can't give a reference because the usage wasn't logged.
The joke is that you would intentionally spell it "ayy lamo" to spot anyone new.

src/ja/common/message/base.py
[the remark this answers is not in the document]
ammen99: nvm, just tried to implement it and saw what you mean... Unfortunate.

src/ja/common/docker_context.py
```diff
@@ -65,6 +149,18 @@ def __init__(self, cpu_threads: int = -1, memory: int = 1):
         @param cpu_threads Initial value of @cpu_threads.
         @param memory The value of @memory.
         """
+        self._cpu_threads = -1
+        if cpu_threads != -1:
```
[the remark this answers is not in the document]
JohannesGaessler: Because the setter checks if you are setting the number of CPU threads to a value > 0.
The current behavior is that you are allowed to set cpu_threads to -1 through the constructor, but not through the setter.

nikolatzotchev:
Can we merge this pls.

JohannesGaessler:
Do we really need type safety for our unit tests?
The whole point of type safety is to prevent unforeseen runtime errors.
This makes no difference for unit tests since runtime errors would cause your unit tests to fail so you would notice immediately.
However, if I wanted to make the unittests type safe I would no longer be able to use the mixin approach that I am using now.
The problem is that when you import a subclass of unittest.TestCase the unit tests of that class will be executed in your test suite as well.
As a consequence inheritance like `JobTest -> SerializableTest -> TestCase` would be a giant pain to implement unless you resort to copy-pasting code.

ammen99:
> Do we really need type safety for our unit tests?
> The whole point of type safety is to prevent unforeseen runtime errors.

I am using an IDE (rather vim + plugins) where type safety helps me use autocomplete and also catch errors while I am typing. Hence I'd like to use type safe code everywhere.

Seeing that the type is wrong can help you realize your mistake faster if the test fails.

> The problem is that when you import a subclass of unittest.TestCase the unit tests of that class will be executed in your test suite as well.

Right, we can make an exception for the mixin class. Just add `# type: ignore` at the top of the python module for the mixin class. (found it here https://github.com/python/mypy/issues/545 and it seems to work on my computer)

PR #56, opened 2020-01-16
Database types & common.work_machine

M1keReck:
- [x] common.work_machine

- [x] database.types

- [x] Tests for all implemented classes

### DIFF
--- a/src/ja/common/work_machine.py
+++ b/src/ja/common/work_machine.py
@@ -1,4 +1,4 @@
-from typing import Dict
+from typing import Dict, cast
 
 from ja.common.message.base import Serializable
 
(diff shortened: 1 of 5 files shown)


M1keReck:
I'll have to add tests for job_entry, other than that, this is ready for review

src/ja/common/work_machine.py
```diff
@@ -15,28 +15,86 @@ def __init__(self, cpu_threads: int, memory: int, swap: int):
         @param memory The amount of RAM, in MB.
         @param swap The amount of swap space, in MB.
         """
+        if cpu_threads < 0:
+            raise ValueError("CPU threads cannot be negative")
+        if memory < 0:
+            raise ValueError("Memory cannot be negative")
+        if swap < 0:
+            raise ValueError("Swap cannot be negative")
```
JohannesGaessler: What is the purpose of these checks?
Judging by the code you added below, any of these properties can become negative, correct?

src/ja/common/work_machine.py
```diff
@@ -15,28 +15,86 @@ def __init__(self, cpu_threads: int, memory: int, swap: int):
...
+    def __iadd__(self, other: ResourceAllocation) -> "ResourceAllocation":
+        self._cpu_threads += other.cpu_threads
+        self._memory += other.memory
+        self._swap += other.swap
+
+        return self
 
     def to_dict(self) -> Dict[str, object]:
-        pass
+        return self.__dict__
```
JohannesGaessler: Please replace this with a manual implementation where you use the properties instead of the object's attributes.

src/ja/server/database/types/work_machine.py
```diff
@@ -46,6 +50,18 @@ def allocate(self,
         @return True if the allocation or the test was successful, False
           otherwise.
         """
+        diff: ResourceAllocation = self._total_resources - allocation
```
JohannesGaessler: Regarding code style:
In my PR I added leading underscores to local references in methods.
Should we add underscores to local references @ammen99 @nikolatzotchev @M1keReck @jooe1 ?

src/ja/server/database/types/work_machine.py
```diff
@@ -93,29 +119,41 @@ def __init__(self,
...
         """
+        return self._state
 
     @property
     def resources(self) -> WorkMachineResources:
         """!
         @return The resources of the work machine, if it is online,
           otherwise None.
         """
+        if self._state == WorkMachineState.ONLINE:
```
JohannesGaessler: A retired work machine can also still have resources.
Retired means that the work machine will not receive new jobs and is pending shutdown but existing jobs will continue to run until they are done.

src/ja/server/database/types/work_machine.py
```diff
@@ -93,29 +119,41 @@ def __init__(self,
...
+            return self._resources
+        return None
 
     @property
     def connection_details(self) -> WorkMachineConnectionDetails:
         """!
         @return The connection details of the work machine, if it is online,
           otherwise None.
         """
+        if self._state == WorkMachineState.ONLINE:
```
JohannesGaessler: Same as with resources.

src/ja/server/database/types/work_machine.py
nikolatzotchev: > _single_leading_underscore: weak "internal use" indicator. E.g. from M import * does not import objects whose names start with an underscore.
That is from pep8, I'm fine with both if someone else has some preferences.

src/ja/common/work_machine.py
ammen99: I think it might be ok for such small classes? I mean, we know what is going to go out of the class. We don't have nesting either.

src/ja/common/work_machine.py
M1keReck: Correct, I made a mistake, those properties can be negative. But cpu_threads and so on can't, so I'll move those checks to the WorkMachineResources class.

ammen99: Hmm, now that I think of it, I don't think it makes sense for a negative resource allocation??

src/ja/common/work_machine.py
JohannesGaessler: One of the consequences of using __dict__ is that underscores get added to property names.
This means that a user writing a config file would have to guess whether to put an underscore in front of a property name unless we were to put underscores in front of all property names (I don't think we should).

src/ja/common/work_machine.py
M1keReck: It doesn't, but it's useful. I updated the code, I moved those checks to WorkMachineResources in database.types.work_machine.
So now, a ResourceAllocation can have negative attributes but a WorkMachineResources instance can't.

src/ja/server/database/types/work_machine.py
M1keReck: Yes we should, leading underscores indicate that a reference is private so that's actually necessary.

JohannesGaessler: >Yes we should, leading underscores indicate that a reference is private so that's actually necessary.

Well, was there perhaps a misunderstanding?
What I meant is that I put underscores in front of all local references, not just object attributes.
So in this case it would be `_diff` instead of `diff`.

M1keReck: There was, we don't need leading underscores for references inside a method because they can't be accessed outside of that method. Object attributes can be accessed outside of their class so an underscore is necessary.

src/ja/server/database/types/work_machine.py
```diff
@@ -57,6 +75,16 @@ def deallocate(self,
...
+        self._free_resources += allocation
+        return True
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, WorkMachineResources):
+            return self._total_resources == other.total_resources and self._free_resources == other.free_resources
+        return False
 
 
 class WorkMachineConnectionDetails:
```
nikolatzotchev: What do we do with this class.

ammen99: It's a TODO, I am not sure what exactly we'll need.

src/ja/common/work_machine.py
```diff
@@ -16,27 +16,84 @@ def __init__(self, cpu_threads: int, memory: int, swap: int):
...
+        ret: bool = self._cpu_threads < 0 or self._memory < 0 or self._swap < 0
+        return ret
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, ResourceAllocation):
+            return self._cpu_threads == other.cpu_threads and self._memory == other.memory and self._swap == other.swap
+        else:
+            return False
+
+    def __sub__(self, other: ResourceAllocation) -> "ResourceAllocation":
```
ammen99: The parameter type should also be in "" otherwise it won't even compile.

JohannesGaessler:
Before you merge this PR, please squash your commits.

ammen99:
Yes, I find it ready to merge, just squash your commits.

M1keReck:
Done, I think

ammen99:
@M1keReck You still haven't squashed the commits together

PR #57, opened 2020-01-16
Unit tests

### LINKED ISSUE #53: Setup up unit tests in our CI

ammen99:
Fixes #53 

~~However, this PR also includes #54 because all we need to do in the CI is to run the `setup.py` file added in #54~~

~~Hence, adding this as a draft until we merge the job implementation.~~

I cherry-picked the setup.py from #54 and implemented a simple test for the Serializable class, so this is more or less independent of #54

### DIFF
--- a/src/test/serializable/base.py
+++ b/src/test/serializable/base.py
@@ -1,0 +1,33 @@
+from ja.common.message.base import Serializable
+from typing import Dict
+from unittest import TestCase
+
+
+class SimpleSerializable(Serializable):
+    def __init__(self, x: int) -> None:
+        self.x = x
+
+    def to_dict(self) -> Dict[str, object]:
+        return {'x': self.x}
+
+    @classmethod
+    def from_dict(cls, property_dict: Dict[str, object]) -> "SimpleSerializable":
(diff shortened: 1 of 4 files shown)


src/test/serializable/base.py
```diff
@@ -0,0 +1,33 @@
...
+        self.x = x
+
+    def to_dict(self) -> Dict[str, object]:
+        return {'x': self.x}
+
+    @classmethod
+    def from_dict(cls, property_dict: Dict[str, object]) -> "SimpleSerializable":
+        assert len(property_dict) == 1
+        assert 'x' in property_dict
+        assert isinstance(property_dict['x'], int)
```
JohannesGaessler: For a trivial unit test like this I'd say it doesn't matter but for more complicated unit tests please use `self.assertTrue` with a meaningful message to make debugging easier.

ammen99: ah right, I'll use it.

ammen99: This is not a `TestCase`, rather a dummy implementation.

PR #58, opened 2020-01-19
User config impl

### LINKED ISSUE #50: Implementation of user config files

nikolatzotchev:
I also implemented SSHConfig, probably should add more test, but I wanted to created the pull request so you can check if I'm doing everything fine. Used Johannes pull request as base.
Fixes #50


src/ja/user/config/query.py
```diff
@@ -12,70 +13,297 @@ class QueryCommandConfig(UserConfig):
...
+                 special_resources: List[List[str]] = None, cpu_threads: Tuple[int, int] = (1, 1),
+                 memory: Tuple[int, int] = (1, 1), before: datetime = None, after: datetime = None,
+                 ssh_config: SSHConfig = None, verbosity: Verbosity = Verbosity.NONE):
+        if cpu_threads[0] > cpu_threads[1] or cpu_threads[0] < 1:
+            raise ValueError("Invalid range for cpu_threads.")
+        if memory[0] > memory[1] or memory[0] < 1:
+            raise ValueError("Invalid range for memory.")
+        if after is not None and before is not None:
+            if after < before:
+                raise ValueError("incorrect datetime")
```
JohannesGaessler: Your error messages are not very helpful to be honest.
Please write more detailed error messages that give more insight into why the parameters are not valid.

src/ja/user/config/query.py
```diff
@@ -12,70 +13,297 @@ class QueryCommandConfig(UserConfig):
     matches any of the provided values.
     """
 
+    def __init__(self, uid: List[str] = None, label: List[str] = None, owner: List[str] = None,
+                 priority: List[JobPriority] = None, status: List[JobStatus] = None, is_preemtible: bool = None,
+                 special_resources: List[List[str]] = None, cpu_threads: Tuple[int, int] = (1, 1),
+                 memory: Tuple[int, int] = (1, 1), before: datetime = None, after: datetime = None,
```
JohannesGaessler: (1, 1) are bad default values for cpu_threads and memory because they are included in the range of valid values.
Please replace them with something like None, (None, None), or (-1, -1).

src/ja/user/config/query.py
```diff
@@ -12,70 +13,297 @@ class QueryCommandConfig(UserConfig):
...
+                   and self._verbosity == o._verbosity
+        else:
+            return False
+
+    def __dir__(self) -> Iterable[str]:
+        return ["_uid", "_label", "_owner", "_priority", "_status", "_is_preemtible",
+                "_special_resources", "_cpu_threads", "_memory", "_before", "_after",
+                "_ssh_config", "_verbosity"]
+
+    def source_from_user_config(self, other_config: "QueryCommandConfig", unset_only: bool = True) -> None:
```
JohannesGaessler: Same as AddCommandConfig.source_from_user_config

src/ja/user/config/query.py
```diff
@@ -12,70 +13,297 @@ class QueryCommandConfig(UserConfig):
...
+        _dict["before"] = self._before.strftime("%d/%m/%Y %H:%M:%S")
+        _dict["after"] = self._after.strftime("%d/%m/%Y %H:%M:%S")
+        _dict["ssh_config"] = self._ssh_config.to_dict()
+        _dict["verbosity"] = self._verbosity
+        return _dict
+
+    @classmethod
+    def from_dict(cls, property_dict: Dict[str, object]) -> "QueryCommandConfig":
+        _uid = None
+        if "uid" in property_dict:
```
JohannesGaessler: These checks should not be necessary.
cls._get_str_from_dict should return None if the dictionary does not contain the specified property.

nikolatzotchev: But it doesn't work with string list.

nikolatzotchev:
I moved the query config and cancel config to the message classes, and also changed the userconfig. I sow that Johannes also implemented sshconfig in his latest pr, so I don't now should I remove mine or wait till his is merged.

src/ja/user/config/base.py
```diff
@@ -19,31 +21,69 @@ class UserConfig(Config):
     Base class for user client Config classes.
     """
 
+    def __init__(self, shh_config: SSHConfig = None, verbosity: Verbosity = None,
+                 config_file: str = str(Path.home()) + "/.config/job_adder/user_config") -> None:
+        if shh_config is None and verbosity is None:
```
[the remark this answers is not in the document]
nikolatzotchev: should I then just remove it?

src/ja/user/message/query.py
```diff
@@ -1,25 +1,262 @@
...
+                cls._raise_error_wrong_type(
+                    key="memory", expected_type="Tuple[int, int]",
+                    actual_type=prop.__class__.__name__
+                )
+            memory = cast(Tuple[int, int], prop)
+
+        before_string = cls._get_str_from_dict(property_dict=property_dict, key="before", mandatory=False)
+        before = None
+        if before_string is not None:
+            before = datetime.strptime(before_string, "%d/%m/%Y %H:%M:%S")
```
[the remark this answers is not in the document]
nikolatzotchev: I can.

src/ja/user/config/add.py
```diff
@@ -7,20 +9,73 @@ class AddCommandConfig(UserConfig):
...
+
     @property
     def job(self) -> Job:
         """!
         @return: The job to be added.
         """
+        return self._job
 
     @property
     def attach(self) -> bool:
```
JohannesGaessler: I think that we discussed this option in the meeting and agreed that we would not implement it.

ammen99: After the meeting :)
But yeah I think we can leave this one, it is not crucial to our project.

nikolatzotchev: I removed it.

src/ja/common/job.py
```diff
@@ -209,10 +208,12 @@ def status(self, new_status: JobStatus) -> None:
         QUEUED -> CANCELLED | RUNNING
         RUNNING -> PAUSED | DONE | CANCELLED | CRASHED
         PAUSED -> CANCELLED | RUNNING
+<<<<<<< HEAD
```
M1keReck: Is this an unfixed merge conflict?

src/ja/common/job.py
```diff
@@ -145,7 +145,6 @@ def uid(self, value: str) -> None:
         """!
         Set the job UID.
         If the job has an UID already set, this results in a RuntimeError.
-
```
ammen99: I don't see why you delete these lines in this file

src/ja/user/config/add.py
```diff
@@ -7,20 +9,75 @@ class AddCommandConfig(UserConfig):
...
+        if isinstance(o, AddCommandConfig):
+            return self._job == o._job \
+                and self._attach == o._attach \
+                and self._blocking == o._blocking \
+                and self._ssh_config == o._ssh_config and self._verbosity == o._verbosity
+        else:
+            return False
+
+    def source_from_add_config(self, other_config: "AddCommandConfig", unset_only: bool = True) -> None:
+        if unset_only is False:
```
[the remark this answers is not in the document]
ammen99: _attach doesn't seem to be handled the same in both cases.
Why not just unify the two cases as I suggested?

src/ja/user/config/add.py
```diff
@@ -30,7 +30,7 @@ def blocking(self) -> bool:
         return self._blocking
 
     def __dir__(self) -> Iterable[str]:
-        return ["_blocking", "_attach", "_job"]
+        return ["_blocking", "_attach", "_job", "_ssh_config"]
```
ammen99: Wasn't there also verbosity?

nikolatzotchev: I was thinking it will be the same for all ad commands but you are right.

Issue #72, opened 2020-01-21
IWorkerProxy


ammen99:
A mock implementation for the proxy that the Server uses to communicate to Workers.

Issue #73, opened 2020-01-21
IProxyFactory


ammen99:
A mock implementation for the ProxyFactory, which returns mock WorkerProxies

PR #84, opened 2020-01-22
Implementation of WorkerProxyDummy, WorkerProxyFactory

JohannesGaessler:
Fixes https://github.com/DistributedTaskScheduling/JobAdder/issues/72 and https://github.com/DistributedTaskScheduling/JobAdder/issues/73 . 
This PR implements a dummy for WorkerProxy and WorkerProxyFactory.

- [x] WorkerProxyDummy
- [x] WorkerProxyFactory

I also implemented Response and WorkerResponse for testing.
@M1keReck you might want to reuse those implementations for your implementation of a dummy for ServerProxy.


JohannesGaessler:
I also did a fixup for Job because I forgot to implement to check whether status transitions are legal.

src/test/proxy/worker_proxy.py
```diff
@@ -0,0 +1,253 @@
...
+from ja.server.proxy.proxy import IWorkerProxy
+from test.abstract import skipIfAbstract
+
+
+class WorkerProxyDummy(IWorkerProxy):
+    def __init__(self, uid: str, jobs: List[Job]):
+        self._uid = uid
+        self._jobs: List[Job] = []
+        for job in jobs:
+            job_copy = Job.from_dict(job.to_dict())
```
ammen99: Why not use deepcopy?

src/test/proxy/worker_proxy_factory.py
```diff
@@ -0,0 +1,44 @@
...
+            factory = WorkerProxyDummyFactory(database=None)
+            wm_wmproxy_tuples: List[Tuple[WorkMachine, IWorkerProxy]] = [
+                (WorkMachine(uid="worker001"), None),
+                (WorkMachine(uid="Worker002"), None),
+                (WorkMachine(uid="Worker003"), None),
+            ]
+            for index in indices:
+                work_machine, saved_proxy = wm_wmproxy_tuples[index]
+                retrieved_proxy = factory.get_proxy(work_machine)
+                self.assertFalse(
```
ammen99: there is assertIsNotNone

src/test/proxy/worker_proxy.py
JohannesGaessler: This isn't just to create a copy but also to simulate the transformation of object -> YAML and YAML -> object that will take place in the real proxy.

src/test/proxy/worker_proxy_factory.py
```diff
@@ -0,0 +1,44 @@
...
+                    retrieved_proxy is None,
+                    "Retrieved proxy for index %s in permutation %s is None." % (index, indices)
+                )
+                if saved_proxy is None:
+                    wm_wmproxy_tuples[index] = work_machine, retrieved_proxy
+                else:
+                    self.assertTrue(
```
ammen99: assertIs ?

src/test/proxy/worker_proxy.py
```diff
@@ -0,0 +1,253 @@
...
+
+    @skipIfAbstract
+    def test_dispatch_same_job_twice_1(self) -> None:
+        response_1 = self._empty_worker.dispatch_job(self._job_1)
+        self.assertTrue(response_1.is_success)
+        response_2 = self._empty_worker.dispatch_job(self._job_1)
+        self.assertFalse(response_2.is_success)
+
+    @skipIfAbstract
+    def test_dispatch_same_job_twice_2(self) -> None:
```
ammen99: this is a duplicate test, I don't think it makes much sense to have it. At least put it in a for loop in the same test ;)

src/test/proxy/worker_proxy.py
```diff
@@ -0,0 +1,253 @@
...
+    @skipIfAbstract
+    def test_cancel_same_job_twice_1(self) -> None:
+        response_1 = self._busy_worker.cancel_job(self._uid_1)
+        self.assertTrue(response_1.is_success)
+        response_2 = self._busy_worker.cancel_job(self._uid_1)
+        self.assertFalse(response_2.is_success)
+
+    @skipIfAbstract
+    def test_cancel_same_job_twice_2(self) -> None:
+        response_1 = self._busy_worker.cancel_job(self._uid_2)
```
ammen99: same as above, duplicate tests.

ammen99: Also the same things below, I know these are tests, but shouldn't we make them more concise?

ammen99: For example, you could iterate over all jobs, and try to pause them on both machines, and expect success depending on whether the job is in proxy._jobs (maybe make that public)

src/test/proxy/worker_proxy.py
```diff
@@ -0,0 +1,253 @@
...
+                        result_string=IWorkerProxy.PAUSE_JOB_SUCCESS % (uid, self.uid), is_success=True).to_dict())
+                except ValueError:
+                    return WorkerResponse.from_dict(WorkerResponse(
+                        result_string=IWorkerProxy.PAUSE_JOB_NOT_RUNNING % (uid, self.uid), is_success=False
+                    ).to_dict())
+        return WorkerResponse.from_dict(WorkerResponse(
+            result_string=IWorkerProxy.PAUSE_JOB_UNKNOWN_JOB % (uid, self.uid), is_success=False).to_dict())
+
+    def resume_job(self, uid: str) -> WorkerResponse:
+        for job in self._jobs:
```
ammen99: not sure how much functional-style python you want to use, this could be `next(job for job in self._jobs if job.uid == uid)`

ammen99: of course same for other functions, although I'm not sure what way is more pythonic

JohannesGaessler: I think what I currently have is easier to understand.

src/test/proxy/worker_proxy.py
ammen99: makes sense

src/test/proxy/worker_proxy.py
JohannesGaessler: This is not a duplicate test because I'm assuring that the position of the job in the list does not affect the test results.

JohannesGaessler: Sorry, I misrememberd how I arranged the jobs.
It's true that this test does not add much value but at the same time it took almost no effort to create.

JohannesGaessler: It is very unlikely that the actual job used makes a difference but I think with unit tests you should test even things that you're relatively certain about.

ammen99: I understand it is easy to add more tests like this, but I think you should also consider what would happen if we need to tweak the interface a little bit - you'd need to make the same changes in a very repetitive manner to fix the tests, hence I suggested grouping tests together with for-loops.

src/test/proxy/worker_proxy.py
JohannesGaessler: I think unit tests should be kept as simple as possible.
You could come up with some approach that took less lines of code but I think that would also be more difficult to debug if any of the tests fail.
If you just have a few lines of code with no if statements and no loops I think that is ideal for a test.

ammen99: I was just saying what I think is standard practice - remove repetitive code as much as possible: https://stackoverflow.com/questions/129693/is-duplicated-code-more-tolerable-in-unit-tests

I would not hold the PR from being merged though, especially not over this.

PR #86, opened 2020-01-22
Database mockup

### LINKED ISSUE #59: Mock database
This is a database which is used for unit tests - it doesn't need to connect to a real database, but should implement the `ServerDatabase` interface.

nikolatzotchev:
Ok, so this sort of works I wanted to show you what I came up with.
There are a few problems, first we cannot have ABC, because there are some metadata problems when having both ABC and BASE. And secondly i haven't tested ob it is going to work when creating a class with from_dict() I will test later or tomorrow, the nice thing using this method is that there is no sql code being written. 

Fixes #59


nikolatzotchev:
Also probably there are going to be some problems with mypy.

src/test/server/database/test_database.py
```diff
@@ -0,0 +1,159 @@
...
+        self.assertTrue(self.mockDatabase.get_jobs_on_machine(self.work_machine) == [self.job])
+
+    def test_schedule(self) -> None:
+        self.job2.status = JobStatus.KILLED
+        self.job2.uid = "job2"
+        self.job.status = JobStatus.RUNNING
+        self.mockDatabase.update_job(self.job)
+        self.mockDatabase.update_job(self.job2)
+        schedule = self.mockDatabase.get_current_schedule()
+        self.assertTrue(schedule, [self.job2])
```
JohannesGaessler: I think the logic for this check is broken.

ammen99:
Another thing I do not like is that right now we have the SQL things all over the database.
Could we perhaps add all these class attributes with `setattr` in the constructor of the SQL database? It is a hack but afaict the whole thing with these queries is not strictly typed internally.

src/ja/server/database/sql/mockup_database.py
```diff
@@ -0,0 +1,121 @@
...
+        self.session_maker: sessionmaker = sessionmaker(self.engine)
+        Base.metadata.create_all(self.engine)
+        self._session = self.session_maker()
+
+    def __del__(self) -> None:
+        self._session.close()
+
+    def find_job_by_id(self, job_id: str) -> Optional[Job]:
+        job: Optional[Job] = self._session.query(Job).filter(Job.job_uid == job_id).first()
+        if job is None:
```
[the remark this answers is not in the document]
nikolatzotchev: yes sorry that was before I added Optional

src/ja/server/database/sql/mockup_database.py
```diff
@@ -0,0 +1,121 @@
...
+from ja.server.database.database import ServerDatabase
+from ja.server.database.types.job_entry import DatabaseJobEntry, JobRuntimeStatistics
+from ja.server.database.types.work_machine import WorkMachine
+from sqlalchemy.orm import sessionmaker
+
+
+class MockDatabase(ServerDatabase):
+    # postgresql = testing.postgresql.Postgresql()
+    # connect to PostgreSQL
+    # engine = create_engine(postgresql.url())
```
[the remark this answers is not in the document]
nikolatzotchev: I removed them now.

src/ja/server/database/sql/base.py
```diff
@@ -0,0 +1,3 @@
+from sqlalchemy.ext.declarative import declarative_base
+
+Base = declarative_base()
```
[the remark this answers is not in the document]
nikolatzotchev: That was the name used in the documentation they provided and that is way I stuck with it. How should I name it.

src/test/server/database/test_database.py
nikolatzotchev: why

nikolatzotchev:
2020-02-05-135918_grim
I don't get why I get this error, I clearly have this status, you changed it in master with your pr but I haven't rebased anything. Maybe I will just rebase.

nikolatzotchev:
@JohannesGaessler I fixed one of your test, because I changed is_preemtible to is_preemptible. It should be is_preemptible right?

src/test/server/database/test_database.py
JohannesGaessler: Because calling self.assertTrue will evaluate schedule as a boolean and [self.job2] as an error message.

JohannesGaessler:
@nikolatzotchev Yes. That is a typo on my part.

src/ja/server/database/sql/base.py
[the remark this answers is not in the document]
JohannesGaessler: If it's the same name as in the documentation I think that changes things.
I think we should keep the name the same in that case.

src/ja/server/database/sql/mock_database.py
```diff
@@ -0,0 +1,118 @@
...
+            self.session.delete(work_machine)
+            self.session.flush()
+            self.session.add(machine)
+        self.session.commit()
+
+    def get_work_machines(self) -> Optional[List[WorkMachine]]:
+        work_machines: Optional[List[WorkMachine]] = self.session.query(WorkMachine).all()
+        return work_machines
+
+    JobDistribution = List[DatabaseJobEntry]
```
ammen99: Use `ServerDatabase.JobDistribution` instead of redeclaring it.

src/ja/server/database/sql/mock_database.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def get_current_schedule(self) -> Optional[JobDistribution]:
+        jobs: Optional[Optional[List[DatabaseJobEntry]]] = self.session.query(DatabaseJobEntry).join(Job) \
```
ammen99: Is the double optional needed here?

ammen99:
Hm, another thing, there is a breaking change (DatabaseJobEntry.stats renamed to statistics), if anybody is using this

src/test/server/database/test_database.py
nikolatzotchev: Thank you I changed every assertTrue with Equal now, didn't now that it was evaluated as boolean

nikolatzotchev:
I renamed the stuff back to the old ones.

ammen99:
@JohannesGaessler You forgot to squash the commits :)

JohannesGaessler:
Yeah, I just noticed that, too.

Issue #74, opened 2020-01-21
Dispatcher


ammen99:
Implementation of the dispatcher class.

PR #87, opened 2020-01-26
Implementation of the dispatcher

JohannesGaessler:
Fixes https://github.com/DistributedTaskScheduling/JobAdder/issues/74 .
I implemented the dispatcher and tested it using my implementation of the factory for worker proxy dummies.
This PR therefore depends on https://github.com/DistributedTaskScheduling/JobAdder/pull/84


src/ja/server/dispatcher/dispatcher.py
```diff
@@ -25,13 +27,26 @@ def set_distribution(self, job_distribution: JobDistribution) -> None:
...
+                        proxy.resume_job(job.uid)
+                    self._previous_statuses[job.uid] = job.status
+                elif job.status == JobStatus.CANCELLED:
+                    proxy.cancel_job(job.uid)  # Do not add job back to self._previous_statuses
+                elif job.status == JobStatus.PAUSED:
+                    proxy.pause_job(job.uid)
+                    self._previous_statuses[job.uid] = job.status
+                elif job.status == JobStatus.QUEUED:
+                    self._previous_statuses[job.uid] = job.status
+                elif job.status in [JobStatus.CRASHED, JobStatus.DONE]:
```
ammen99: `CRASHED` and `DONE` should never make it to the dispatcher.
The scheduler does not concern itself with those states at all.

ammen99: This also means that the dispatcher should clear its previous status for any job which is not present in the new distribution.

src/test/server/test_dispatcher.py
```diff
@@ -0,0 +1,125 @@
...
+from ja.server.database.types.job_entry import DatabaseJobEntry
+from ja.server.database.types.work_machine import WorkMachine
+from ja.server.dispatcher.dispatcher import Dispatcher
+from test.proxy.worker_proxy_factory import WorkerProxyDummyFactory
+
+
+class TestDispatcher(TestCase):
+
+    def _assert_distribution_correct_a(self) -> None:
+        # Test if jobs are dispatched correctly by invoking the proxies:
```
ammen99: Why do we need to test this in an indirect way here?
We control the factory and we know it returns mock objects. We can directly access those mock objects and check the job status, right?

JohannesGaessler: I implemented tests using only the public interface so that we can directly compare the results of mock implementations and the actual implementations during the integration phase.

ammen99:
Needs a rebase btw.

PR #88, opened 2020-01-27
worker config #63

### LINKED ISSUE #63: Worker config
Worker configuration classes, it should be able to read for ex. the IP address of the server.

jooe1:
Fixes #63

### DIFF
--- a/src/ja/worker/config.py
+++ b/src/ja/worker/config.py
@@ -2,8 +2,6 @@ from typing import Dict
 from ja.common.config import Config
 from ja.common.proxy.ssh import SSHConfig
 from ja.common.work_machine import ResourceAllocation
-import yaml
-from typing import cast
 
 
 class WorkerConfig(Config):
--- a/src/ja/worker/config.py
+++ b/src/ja/worker/config.py
@@ -9,7 +9,7 @@ class WorkerConfig(Config):
     Config for the worker client.
     """
 
-    def __init__(self, uid: str, ssh_config: SSHConfig, resource_allocation: ResourceAllocation):
+    def __init__(self, uid: str = None, ssh_config: SSHConfig = None, resource_allocation: ResourceAllocation = None):
         """!
         creates resource allocation instance
         """
--- a/src/ja/worker/config.py
+++ b/src/ja/worker/config.py
@@ -1,12 +1,9 @@
 from typing import Dict
-
-from ..common.config import Config
-from ..common.proxy.ssh import SSHConfig
-from ..common.work_machine import ResourceAllocation
+from ja.common.config import Config
+from ja.common.proxy.ssh import SSHConfig
+from ja.common.work_machine import ResourceAllocation
 import yaml
 import os
-import psutil
-
 
 
 class WorkerConfig(Config):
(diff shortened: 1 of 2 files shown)


src/ja/worker/config.py
```diff
@@ -1,36 +1,76 @@
...
+    def __init__(self, conf_path: str):
+        """!
+        reads the config file from the disk
+        creates resource allocation instance
+        """
+        self._conf_path = conf_path
+        self._ssh_config = self.read_conf()
+        self._resource_allocation = self.get_resources()
+
+    def read_conf(self) -> SSHConfig:
```
JohannesGaessler: This method is not needed.
YAML files can be automatically converted into Python dictionaries.
We can therefore just use the from_dict method to convert a YAML file to a Python object.

src/ja/worker/config.py
```diff
@@ -1,36 +1,76 @@
...
+        with open(self._conf_path, 'r') as stream:
+            data_loaded = yaml.safe_load(stream)
+
+        hostname: str = data_loaded["hostname"]
+        username: str = data_loaded["username"]
+        self._uid = data_loaded["uid"]
+        return SSHConfig(hostname, username)
+
+    def get_resources(self) -> ResourceAllocation:
```
JohannesGaessler: We do not need this method.
The amount of resources should be read from the configuration file instead of being determined automatically.
This is because the administrator might want to reserve some amount of resources for other processes.

ammen99: Maybe we can have this as a fallback? If nothing is specified in the config file.

JohannesGaessler: I think we should not provide a fallback at all.
I think the administrator should be forced to make a decision regarding how many resources the worker client should be using because we cannot provide a default that would work well in all situations.

src/ja/worker/config.py
```diff
@@ -1,36 +1,68 @@
...
+            data_loaded = yaml.safe_load(stream)
+
+        self._uid = data_loaded["uid"]
+        hostname: str = data_loaded["hostname"]
+        username: str = data_loaded["username"]
+        self._ssh_config = SSHConfig(hostname, username)
+        cpu_threads: int = cast(int, data_loaded["cpu_threads"])
+        available_memory: int = cast(int, data_loaded["available_memory"])
+        swap_memory: int = cast(int, data_loaded["swap_memory"])
+        self._resource_allocation = ResourceAllocation(cpu_threads, available_memory, swap_memory)
```
JohannesGaessler: I think you misunderstood what I meant.
You should read in properties via from_dict, not via the constructor.
The YAML file should be opened in the main class of the worker, not in WorkerConfig.

JohannesGaessler:
I pushed an implementation of SSHConfig that you can use: https://github.com/DistributedTaskScheduling/JobAdder/pull/93

JohannesGaessler:
>Also, we should have an option for the server IP address the client needs to connect to.

What do you mean by this?
SSHConfig has a hostname property that determines the target for SSH connections.

ammen99:
> What do you mean by this?
> SSHConfig has a hostname property that determines the target for SSH connections.

Ah sorry, it is not needed then.

Issue #67, opened 2020-01-21
IServerProxy


ammen99:
The proxy which the user client uses to communicate with the server.
This is just about a mock implementation, which doesn't really send the commands over the network, but is used only in unit tests.

PR #89, opened 2020-01-27
User CLI

### LINKED ISSUE #68: CLI Handler
The handling of command line arguments by the user client.

M1keReck:
Fixes #68 and #67

- [x] ServerProxy

- [x] Add command

- [x] Query command

- [x] Cancel command

- [x] Tests


src/test/proxy/server_proxy.py
```diff
@@ -0,0 +1,44 @@
...
+from ja.common.job import Job, JobStatus
+
+from ja.user.config.add import AddCommandConfig
+from ja.user.config.cancel import CancelCommandConfig
+from ja.user.config.query import QueryCommandConfig
+
+
+class ServerProxyDummy(IUserServerProxy):
+
+    def __init__(self, db: ServerDatabase)
```
JohannesGaessler: I think that the dummy for ServerProxy should not depend on ServerDatabase.
The point of the dummy is to implement and test the other components of the user client without needing an implementation of the server or its components, right?
@ammen99 @nikolatzotchev what do you think?

src/ja/user/cli.py
```diff
@@ -1,19 +1,113 @@
...
+            #not sure what to put here yet
+
+        #parser for cancel command
+        parser_cancel = subparsers.add_parser("cancel")
+        parser_cancel.add_argument("--name", "-n", required=True, help="the name of the job that needs to be cancelled")
+
+        args: Namespace = parser.parse_args()
+
+        if args.command == "add":
+            ssh: SSHConfig = SSHConfig(os.getuid())
```
JohannesGaessler: Calling os.getuid() on the user client is pointless.
What matters is the user id on the server.
Also, SSHConfig doesn't even have uid as a property, right?

src/test/proxy/server_proxy.py
nikolatzotchev: Yes, dummy is supposed to be something simple, that is used for testing other components. The database is not really useful here.

src/ja/user/cli.py
```diff
@@ -1,19 +1,113 @@
...
+        parser_add.add_argument("--threads", "-t", type=int, default=threads, help="Assign a number of cpu threads required for your job.")
+        parser_add.add_argument("--ram", "--memory", "-r", type=int, default=ram, help="Assign a specific amount of ram to your job")
+        parser_add.add_argument("--preemptible", "--pe", action="store_true", default=preemptible)
+        parser_add.add_argument("--mount", "-m", nargs="+", type=str, default=mount)
+        parser_add.add_argument("--email", "-e", required=email is None, default=email)
+        
+        #parser for query command
+        parser_query = subparsers.add_parser("query")
+        parser_query.add_argument("--name", "--uid", "-n", dest="name", required=True, )
+            #not sure what to put here yet
```
JohannesGaessler: You're supposed to put one CLI argument for each property of QueryCommand(Config) here.

src/test/proxy/server_proxy.py
ammen99: Actually my idea of a Mock database was to be able to use it everywhere where applicable, even here if it simplifies things.

However, I do not think that it would take much effort to actually store a list of jobs instead of the whole database.

src/ja/user/cli.py
M1keReck: SSHConfig has username and hostname as mandatory parameters in the constructor, I thought username refers to the local user id.
How should I initiate a SSHConfig instance?

JohannesGaessler: username is the name of the user as a string that you would use for login via ssh.
uid is an int that is used on Linux systems as an identifier for users.
The parameters for SSHConfig should be read from command line arguments and/or the default config.

JohannesGaessler: I'm currently implementing SSHConnection and I will soon push a branch that implements SSHConfig.
You could rebase your PR onto that branch to save yourself the work of implementing SSHConfig.

src/test/proxy/server_proxy.py
```diff
@@ -0,0 +1,44 @@
...
+        self._db = db
+
+    def add_job(self, add_config: AddCommandConfig) -> ServerResponse:
+        job: Job = add_config.job
+        db_job: Job = self._db.find_job_by_id(job.uid)
+        if db_job is not None:
+            return ServerResponse(result_string = "Job with id %s already exists" % ,
```
JohannesGaessler: I think this error message is broken because you have nothing after the percent sign.

JohannesGaessler:
I pushed an implementation of SSHConfig that you can use: https://github.com/DistributedTaskScheduling/JobAdder/pull/93

src/ja/user/cli.py
```diff
@@ -48,66 +62,140 @@ def get_server_command(self) -> ServerCommand:
...
+                                  help="Jobs scheduled before this point in time (YYYY-MM-DD[ HH:MM:SS]).")
+        parser_query.add_argument("--after",
+                                  help="Jobs scheduled after this point in time (YYYY-MM-DD[ HH:MM:SS]).")
+
+        # parser for cancel command
         parser_cancel = subparsers.add_parser("cancel")
-        parser_cancel.add_argument("--name", "-n", required=True, help="the name of the job that needs to be cancelled")
+        parser_cancel.add_argument("label_or_uid",
+                                   choices=["label", "uid"],
+                                   help="You can cancel a job by specifying it's uid or it's label")
```
JohannesGaessler: it's -> its

M1keReck:
I am using a lot of classes that haven't been merged yet, so I am going to postpone tests and move to another issue.

ammen99:
Instead of merging the master branch, can you rebase on top of it? This way we can see only the changes made in this PR, which makes it easier to review.

M1keReck:
@ammen99 I just created another branch since that appeared to be easier, I'll close this PR

PR #90, opened 2020-01-27
Default algorithm implementation

### LINKED ISSUE #79: Default scheduling algorithm
Uses #78

ammen99:
Also contains my other work from #85 
TODO:

- [x] Implement blocking jobs

- [x] Urgent job on blocked machine test

- [x] Order of consideration: paused jobs before all other non-preempting jobs

- [x] Investigate why caller WorkMachines is modified even when the algorithm takes a deepcopy

Fixes #79

### DIFF
--- a/src/ja/server/scheduler/default_algorithm.py
+++ b/src/ja/server/scheduler/default_algorithm.py
@@ -1,7 +1,21 @@
-from typing import List
+from copy import deepcopy
+from ja.common.job import JobStatus, Job
+from ja.common.work_machine import ResourceAllocation
 from ja.server.database.database import ServerDatabase
+from ja.server.database.types.job_entry import DatabaseJobEntry, JobRuntimeStatistics
 from ja.server.database.types.work_machine import WorkMachine
 from ja.server.scheduler.algorithm import SchedulingAlgorithm, JobDistributionPolicy, CostFunction
+from typing import List, Dict, Tuple
+
+
+def get_allocation_for_job(job: Job, other_status: JobStatus = None) -> ResourceAllocation:
+    """
+    Get the resource allocation for a job depending on its status.
+    """
+    status = other_status if other_status else job.status
+    if status is JobStatus.PAUSED:
+        return ResourceAllocation(0, 0, job.docker_constraints.memory)
+    return ResourceAllocation(job.docker_constraints.cpu_threads, job.docker_constraints.memory, 0)
 
 
 class DefaultSchedulingAlgorithm(SchedulingAlgorithm):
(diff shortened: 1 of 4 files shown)


ammen99:
This PR is now ready for review, but should probably be reviewed after #85

ammen99:
Added support for special resources as well. They are checked before any of the policies are invoked.

src/ja/server/scheduler/default_algorithm.py
```diff
@@ -17,16 +31,150 @@ def __init__(self,
...
+        if not preemptive:
+            return False
+
+        (machine, preempted_jobs) = preemptive
+        for preempt in preempted_jobs:
+            self._set_state(preempt, next_schedule, machine, JobStatus.PAUSED)
+        self._set_state(job.job, next_schedule, machine, JobStatus.RUNNING)
+        return True
+
+    def _deepcopy_args(self, current_schedule: ServerDatabase.JobDistribution, available_machines: List[WorkMachine]) \
```
nikolatzotchev: you can add *@ staticmethod* to this if you want

src/ja/server/scheduler/default_algorithm.py
```diff
@@ -17,16 +31,150 @@ def __init__(self,
...
+    def _set_state(self,
+                   job: Job,
+                   schedule: ServerDatabase.JobDistribution,
+                   machine: WorkMachine,
+                   new_status: JobStatus) -> None:
+        if job.status is JobStatus.RUNNING or job.status is JobStatus.PAUSED:
+            machine.resources.deallocate(get_allocation_for_job(job))
+
+        job.status = new_status
+        machine.resources.allocate(get_allocation_for_job(job))
```
M1keReck: This function only works for some status changes since it wouldn't work with crashed or cancelled as a new status right?

ammen99: Yes, crashed or cancelled is not handled by the scheduling algorithm.

src/ja/server/scheduler/default_algorithm.py
```diff
@@ -17,16 +31,150 @@ def __init__(self,
...
+        self._preemptive_policy = preemptive_distribution_policy
+        self._reserved_machines: Dict[str, str] = {}  # Job UID -> Machine UID
+        self._cost_cache: Dict[str, float] = {}  # Job UID -> Effective cost
+
+    def _set_state(self,
+                   job: Job,
+                   schedule: ServerDatabase.JobDistribution,
+                   machine: WorkMachine,
+                   new_status: JobStatus) -> None:
+        if job.status is JobStatus.RUNNING or job.status is JobStatus.PAUSED:
```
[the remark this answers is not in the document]
ammen99: yep, what you suggest is more readable.

src/ja/server/scheduler/default_algorithm.py
ammen99: ok, done.

PR #91, opened 2020-01-29
iworker-server-proxy #64

### LINKED ISSUE #64: IWorkerServerProxy
A mock implementation for the Proxy which the Worker uses to communicate with the Server. This class can then be used to test that for ex. the DockerEvents class correctly communicates with the Server that a job has finished/crashed, etc.

jooe1:
Fixes #64

### DIFF
--- a/src/ja/worker/proxy/proxy.py
+++ b/src/ja/worker/proxy/proxy.py
@@ -4,6 +4,11 @@ from ja.common.proxy.proxy import ContinuousProxy
 from ja.common.message.server import ServerResponse
 from ja.common.proxy.ssh import SSHConfig
 from ja.server.database.types.work_machine import WorkMachineResources
+from ja.worker.message.register import RegisterWorkerCommand
+from ja.worker.message.retire import RetireWorkerCommand
+from ja.worker.message.done import JobDoneCommand
+from ja.worker.message.crashed import JobCrashedCommand
+from ja.server.database.database import ServerDatabase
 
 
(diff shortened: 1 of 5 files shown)


src/ja/worker/proxy/proxy.py
```diff
@@ -4,6 +4,11 @@
...
 from ja.common.proxy.ssh import SSHConfig
 from ja.server.database.types.work_machine import WorkMachineResources
+from ja.worker.message.register import RegisterWorkerCommand
+from ja.worker.message.retire import RetireWorkerCommand
+from ja.worker.message.done import JobDoneCommand
+from ja.worker.message.crashed import JobCrashedCommand
+from ja.server.database.database import ServerDatabase
 
 
 class IWorkerServerProxy(ContinuousProxy, ABC):
```
JohannesGaessler: You are not supposed to implement the actual proxy class.
You are supposed to create a proxy dummy in the test package that mimics the behavior of the actual server without being functional.
This dummy must **not** depend on the implementation of the database.
It should keep track of jobs with something like a dictionary instead.

ammen99:
@M1keReck Why was this closed?

M1keReck:
Because @jooe1 told me to, now that I read the PR, I am not sure anymore. I'll let him decide

jooe1:
This was the old iworker server proxy dummy PR and I forgot to close it. I thought i had to do the actual implementation instead of the dummy

ammen99:
Okay so it seems like this should be closed as @jooe1 said.

Issue #76, opened 2020-01-21
Remote


ammen99:
Implementation of the Remote program. Not sure how to unit test this properly, but probably we can extract the Job owner verification and unit test that.

Issue #75, opened 2020-01-21
SSHConnection implementation (+Mock)


ammen99:
Class which does the actual message transmission and calls the Remote on the remote machine.

PR #93, opened 2020-01-31
Implementation of SSHConfig, SSHConnection, Remote

JohannesGaessler:
Fixes https://github.com/DistributedTaskScheduling/JobAdder/issues/75 and fixes https://github.com/DistributedTaskScheduling/JobAdder/issues/76 .

This PR implements:

- [x] SSHConfig
- [x] SSHConnection (no unit tests)
- [x] SSHConnectionDummy
- [x] Remote
- [x] CommandHandler


JohannesGaessler:
I rebased this PR onto https://github.com/DistributedTaskScheduling/JobAdder/pull/87 because both worked on message classes.
Strictly speaking I've implemented all of the classes I've set out to implement but I will probably push another commit to revise the interfaces.

JohannesGaessler:
Unless there turn out to be problems I would consider this implementation to be now complete.

Also, I think that we should eventually remove ServerResponse and WorkerResponse.
But since we have so many PRs open as of right now we should wait with that.

src/test/remote/test_remote.py
```diff
@@ -0,0 +1,54 @@
+import yaml
+from subprocess import run, PIPE
+from unittest import TestCase
+from getpass import getuser
+from typing import Dict
+
+from ja.common.proxy.command_handler import CommandHandler
+
+
+class CommandHandlerDummy(CommandHandler):
```
nikolatzotchev: This is more like a question, isn't it better to have on CommandHandlerDummy, or do you thing it is better to test the way you did.

src/ja/common/message/base.py
```diff
@@ -144,13 +144,16 @@ class Response(Message):
     A base class for messages which are sent as a response to Commands. They
     indicate the result of the action on the remote component.
     """
-    def __init__(self, result_string: str, is_success: bool):
+    def __init__(self, result_string: str, is_success: bool, uid: str = None):
```
ammen99: You should use a subclass here, not add a field which is not relevant for all responses.

src/ja/common/proxy/command_handler.py
```diff
@@ -17,7 +22,48 @@ class CommandHandler(ABC):
...
     and the Command was executed without error, and is false otherwise.
     """
-    def __init__(self, socket: str):
+    def __init__(self, socket_path: str):
         """!
-        @param socket: The Unix named socket to listen for Commands on.
+        @param socket_path: The Unix named socket to listen for Commands on.
         """
+        self._socket_path = socket_path
+        self._listen_thread = Thread(target=self._listen)
```
ammen99: I thought the main loop of the server/worker is going to be the command handler, so no need for additional threads here?

src/ja/common/proxy/command_handler.py
```diff
@@ -17,7 +22,48 @@ class CommandHandler(ABC):
...
+        self._listen_thread.daemon = True  # Ensures that the thread is killed when main thread exits
+        self._listen_thread.start()
+
+    def _listen(self) -> None:
+        # Make sure the socket does not already exist
+        try:
+            os.unlink(self._socket_path)
+        except OSError:
+            if os.path.exists(self._socket_path):
```
ammen99: What happens otherwise? Print an error on the stdout?

src/ja/common/proxy/command_handler.py
```diff
@@ -17,7 +22,48 @@ class CommandHandler(ABC):
...
+        named_socket.bind(self._socket_path)
+        named_socket.listen(1)
+
+        while True:
+            connection, client_address = named_socket.accept()
+            try:
+                input_string = ""
+                while True:
+                    data = connection.recv(1024)
+                    if data[-1:] == b"\0":
```
ammen99: A cleaner solution would be to always interpret the first 4 bytes or so as the length of the message, then reading that much bytes.

src/ja/common/proxy/command_handler.py
```diff
@@ -17,7 +22,48 @@ class CommandHandler(ABC):
     success property of the Response is True if both validations were passed
     and the Command was executed without error, and is false otherwise.
     """
-    def __init__(self, socket: str):
+    def __init__(self, socket_path: str):
```
ammen99: We should probably add the ServerDatabase as parameter here, otherwise you can't call the `execute` methods ...

src/test/remote/test_remote.py
ammen99: I assume the intention is to test the real code of the CommandHandler as well

src/test/remote/test_remote.py
```diff
@@ -0,0 +1,54 @@
...
+        while command_string.startswith("COMMAND"):
+            command_string = command_string[7:]
+            response_string += "RESPONSE"
+        return dict(payload=response_string)
+
+
+class RemoteTest(TestCase):
+    """
+    Class for testing Remote and CommandHandler.
+    IMPORTANT: THESE TESTS WILL FAIL IF YOU DON'T RUN THEM FROM THE BASE SRC DIRECTORY.
```
ammen99: Would this be true if PYTHONPATH is set appropriately??

src/test/remote/test_remote.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+
+    def _call_remote(self, command_string: str) -> str:
+        command_dict = dict(command=dict(payload=command_string), type_name=self._command_string)
+        finished_process = run(
+            [
+                "python3",
+                "test/remote/remote"  # Change to "remote" to run from current directory
```
ammen99: Could we have something like `python3 -m test.remote.remote` and require to just set `PYTHONPATH` ?

src/ja/common/proxy/command_handler.py
```diff
@@ -17,7 +22,48 @@ class CommandHandler(ABC):
...
+                )
+                response_string = yaml.dump(response_dict)
+                connection.sendall(response_string.encode())
+            finally:
+                connection.close()
+
+    @abstractmethod
+    def _process_command_string(
+            self, command_dict: Dict[str, object], type_name: str, username: str) -> Dict[str, object]:
+        pass
```
ammen99: (Not an issue, just thinking) at which point are we going to verify the user's identity? Are we going to pass the username to the `execute` methods and then let the commands check the user?

src/ja/common/message/base.py
JohannesGaessler: I am intentionally not using sub-classes of Response.
Due to the Liskov substitution principle it is not easily possible to return a more specific class when a method is overridden.
I think the implementation will be cleaner this way.

src/ja/common/proxy/command_handler.py
JohannesGaessler: You can just add an argument for the server database for a sub-class constructor.

src/ja/common/proxy/command_handler.py
JohannesGaessler: I think you may be right, but I think a design like this works better.
We're going to have multiple components that need their own thread so I think we should not put the main thread into any of them.

src/ja/common/proxy/command_handler.py
JohannesGaessler: If the deleting the file under _socket_path fails because it does not exist noting is done.

src/ja/common/proxy/command_handler.py
JohannesGaessler: I think checking for permissions should be done in _process_command_string.

src/test/remote/test_remote.py
JohannesGaessler: @nikolatzotchev I don't understand the question.

src/test/remote/test_remote.py
JohannesGaessler: Yes, because PYTHONPATH determines where to look for modules only.
PYTHONPATH does **not** change the working directory of the python process.

src/ja/common/proxy/command_handler.py
ammen99: so what is the main thread going to do then?

JohannesGaessler: Wait, what did we agree on for starting/stopping the server again?
Didn't we need the main thread for that?

src/ja/common/proxy/command_handler.py
JohannesGaessler: I agree.
I pushed a fixup.

src/test/remote/test_remote.py
JohannesGaessler: Wait, I think I understand what you mean.
I'll see if it works.

JohannesGaessler: I pushed a fixup.
Thanks for the pointer.

src/test/remote/test_remote.py
nikolatzotchev: I was confused because you had 2 different commandHandlerDummys, I think I get it now

src/ja/common/message/base.py
ammen99: What? LSP states that you can return with stronger guarantees. You cannot take a more specific argument.

src/ja/common/proxy/command_handler.py
ammen99: Ah yes, I forgot this is a generic class.

src/ja/common/proxy/command_handler.py
```diff
@@ -17,7 +22,45 @@ class CommandHandler(ABC):
...
+                raise
+        named_socket = socket.socket(family=socket.AF_UNIX, type=socket.SOCK_STREAM)
+        named_socket.bind(self._socket_path)
+        named_socket.listen(1)
+
+        while True:
+            connection, client_address = named_socket.accept()
+            try:
+                # First 8 bytes encode command length:
+                command_length = int.from_bytes(bytes=connection.recv(8), byteorder="big")
```
ammen99: Is recv guaranteed to give us 8 bytes? As far as I understand the documentation, you could also get 2 bytes and the next 6 bytes later ...

src/ja/common/proxy/command_handler.py
ammen99: Ah I see.

src/ja/common/proxy/command_handler.py
ammen99: I am not sure we said anything about starting/stopping it.
How is it going to be done?

ammen99: Or maybe we could discuss this in the meeting today.

src/ja/common/message/base.py
JohannesGaessler: I think you're right in that I used the LSP incorrectly.
The specific problem I'm having:
In SSHConnection I want to implement a method like `send_command(command: Command) -> Response`.
If only some sub-classes of Response had the uid property I would need to use casts to access the property.

src/ja/common/proxy/command_handler.py
JohannesGaessler: I think we said that we would make it a systemd service.
We would not need the main thread to control the shutdown in that case, right?

src/ja/common/proxy/command_handler.py
JohannesGaessler: Well, you also have no guarantee to ever receive a special character that signals the end of the input or that the socket will ever be closed.

ammen99: Well I think we will receive everything since it is our own remote that is sending us those messages (so we shouldn't get malicious input)

But the problem is rather, the system is free to send it in chunks of 2 bytes, and we won't handle that (I imagine you could write a small function to read N bytes, which does something like `len(buffer) < N: buffer += recv(N-len(buffer))`

src/ja/common/message/base.py
ammen99: I think a cast should be fine in this case, but whatever, its not critical.

src/ja/common/proxy/command_handler.py
ammen99: So we can just do a while True: sleep(1000) in the main loop .. fine by me.

src/ja/common/proxy/command_handler.py
```diff
@@ -45,20 +45,28 @@ def _listen(self) -> None:
...
-                while True:
-                    data = connection.recv(1024)
-                    if data[-1:] == b"\0":
-                        input_string += data[:-1].decode()
-                        break
-                    input_string += data.decode()
+                command_bytes = bytes(0)  # First 8 bytes encode command length
+                command_length = None
+                while command_length is None or len(command_bytes) < command_length:
+                    command_bytes += connection.recv(1024)
```
ammen99: I think it would have been much simpler if you just read the first 8 bytes with connection.recv(8-len(received)) and then another loop for the other bytes but that's inconsequential.

ammen99:
Hey, the checks weren't passing :)

ammen99:
I removed the changes from master, please fix the issues and open another PR

JohannesGaessler:
Thank you.
I might be misremembering, but didn't we configure this repo so that PRs can only be merged if the travis build succeeds?

ammen99:
I thought so, and I was surprised you have been able to merge it. Maybe 2 approvals override checks or something?

This is enabled for the master branch:
slurped

PR #94, opened 2020-02-02
iworkerServerProxy dummy

jooe1:


src/test/worker/worker_server_proxy.py
```diff
@@ -0,0 +1,118 @@
...
+class WorkerServerProxyDummy(IWorkerServerProxy):
+    def __init__(self, wmcs: Dict[str, WorkMachine], jobs: Dict[str, Job]):
+        self._wmcs: Dict[str, WorkMachine] = copy.deepcopy(wmcs)
+        self._jobs: Dict[str, Job] = copy.deepcopy(jobs)
+
+    def register_self(self, uid: str, work_machine_resources: WorkMachineResources) -> ServerResponse:
+        if uid in self._wmcs:
+            return ServerResponse("work machine already exist!", False)
+        self._wmcs[uid] = WorkMachine(
+            uid="workmachine-1",
```
JohannesGaessler: This should use the uid argument instead of a string literal.

jooe1: the code seem to have some assert false tests, do you have some specific params you want me to check?

src/ja/worker/proxy/proxy.py
```diff
@@ -10,28 +10,30 @@ class IWorkerServerProxy(ContinuousProxy, ABC):
     """
     Interface for the proxy for the central server used on the worker client.
     """
-    def __init__(self, ssh_config: SSHConfig):
+    def __init__(self, uid: str, ssh_config: SSHConfig):
```
nikolatzotchev: why do we need the uid here

src/test/worker/worker_server_proxy.py
JohannesGaessler: Sorry, it seems I made a mistake when reviewing your PR.
It seems your PR already has the tests that I wanted and I just didn't see them.

src/test/worker/worker_server_proxy.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def unregister_self(self, uid: str) -> ServerResponse:
+        if uid in self._wmcs:
+            self._wmcs[uid].state = WorkMachineState.RETIRED
+            return ServerResponse("work machine unregistred!", True)
+        return ServerResponse("work machine does not exist!", False)
+
+    def notify_job_finished(self, uid: str) -> ServerResponse:
+        if uid in self._jobs:
+            self._jobs[uid].status = JobStatus.DONE
```
JohannesGaessler: This will raise a ValueError if the method is called twice for the same Job.
A test for this would be nice to have, but not required.

src/test/worker/worker_server_proxy.py
```diff
@@ -0,0 +1,118 @@
...
+            return ServerResponse("job finished!", True)
+        return ServerResponse("job does not exist!", False)
+
+    def notify_job_crashed(self, uid: str) -> ServerResponse:
+        if uid in self._jobs:
+            self._jobs[uid].status = JobStatus.CRASHED
```
JohannesGaessler: Same as with notify_job_finished.

src/ja/worker/proxy/proxy.py
nikolatzotchev: the commands are called from the worker main, what is the point of having this uid if you don't use it in the commands?

jooe1: I think that was a mistake. I honestly don't remember why i put it there.

PR #97, opened 2020-02-09
Implementation of WorkerProxy

### LINKED ISSUE #77: WorkerProxy
The real proxy the Server uses for communication with the Worker.

JohannesGaessler:
Implementation of WorkerProxy.
~~Based on https://github.com/DistributedTaskScheduling/JobAdder/pull/93 because I needed the SSH classes.~~

Fixes #77

### DIFF
--- a/src/ja/server/proxy/proxy.py
+++ b/src/ja/server/proxy/proxy.py
@@ -1,8 +1,12 @@
 from abc import ABC, abstractmethod
 
 from ja.common.proxy.proxy import ContinuousProxy
-from ja.common.proxy.ssh import SSHConfig
+from ja.common.proxy.ssh import SSHConfig, ISSHConnection, SSHConnection
 from ja.common.message.worker import WorkerResponse
+from ja.common.message.worker_commands.cancel_job import CancelJobCommand
+from ja.common.message.worker_commands.pause_job import PauseJobCommand
+from ja.common.message.worker_commands.resume_job import ResumeJobCommand
+from ja.common.message.worker_commands.start_job import StartJobCommand
 from ja.common.job import Job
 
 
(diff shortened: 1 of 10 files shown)


src/test/proxy/worker_proxy_factory.py
```diff
@@ -42,3 +41,35 @@ def test_retrieval(self) -> None:
...
+
+class WorkerProxyDummyFactoryTest(AbstractWorkerProxyFactoryTest):
+    """
+    Class for testing WorkerProxyDummyFactory.
+    """
+    def _get_factory(self) -> WorkerProxyFactoryBase:
+        return WorkerProxyDummyFactory(database=None)
+
+
+class WorkerProxyFactory(WorkerProxyFactoryBase):
```
ammen99: I find the name confusing because it is the same name as the actual `WorkerProxyFactory`. Can we name it `Mock...` or `...NoCache` or something like that?

src/ja/server/proxy/proxy.py
```diff
@@ -63,23 +51,49 @@ def resume_job(self, uid: str) -> WorkerResponse:
...
+        return response
+
+
+class WorkerProxy(WorkerProxyBase):
+    """
+    Implementation of the proxy for the worker client used on the central server.
+    """
+
+    def _get_ssh_connection(self, ssh_config: SSHConfig) -> ISSHConnection:
+        raise NotImplementedError("Remote path not defined")
```
ammen99: Wouldn't it make sense to put the remote in `PATH` or run it with `python3 -m ...` as I suggested on the other PR? This way you don't need to hardcode the path to it.

src/test/proxy/worker_proxy.py
```diff
@@ -1,253 +1,63 @@
...
+from ja.common.message.worker_commands.start_job import StartJobCommand
+from ja.common.proxy.command_handler import CommandHandler
+from ja.common.proxy.ssh import SSHConfig, ISSHConnection
+from ja.server.proxy.proxy import WorkerProxyBase
+from test.proxy.abstract import AbstractWorkerProxyTest
+from test.proxy.worker_proxy_dummy import WorkerProxyDummy
+from test.ssh.test_ssh_dummy import SSHConnectionDummy
+
+
+class WorkerProxy(WorkerProxyBase):
```
ammen99: Maybe WorkerProxyDummySSH ?

src/test/proxy/worker_proxy.py
```diff
@@ -1,253 +1,63 @@
...
+class WorkerCommandHandlerDummy(CommandHandler):
+
+    def __init__(self, socket_path: str, jobs: List[Job]):
+        super().__init__(socket_path=socket_path)
+        self._worker_proxy_dummy = WorkerProxyDummy(uid=None, jobs=jobs)
+
+    def _process_command_string(
+            self, command_dict: Dict[str, object], type_name: str, username: str) -> Dict[str, object]:
+        assert username == getuser()
+        if type_name == "StartJobCommand":
```
ammen99: To avoid these, can we not dynamically import modules and classes based on their name?

JohannesGaessler: That sounds like it would be more work because you would need to get MyPy to accept it.

ammen99: Just cast?

ammen99:
LGTM. Not giving an approve because I don't want to accidentally merge this before we merge the previous PR.

ammen99:
Needs a rebase.

PR #98, opened 2020-02-09
Implementation of DockerInterface

JohannesGaessler:
I started working on what is supposed to be the docker context.
However, I think that the easiest way to monitor the docker events is to track the mapping of job UIDs to container IDs.
Because of this I've started implementing a DockerInterface class instead which also handles adding, cancelling, pausing, and resuming jobs.


JohannesGaessler:
To properly fix the travis checks I need https://github.com/DistributedTaskScheduling/JobAdder/pull/94 to be merged.

PR #101, opened 2020-02-10
Imlementing SQLDatabase.

nikolatzotchev:
Moved the methods to SQLDatabase, added the callbacks, probably not right but ilia can check. I don't now what kind of test I should write about this. The methods are already tested in the mock_database so no point in testing them, but maybe the callbacks?

### DIFF
--- a/src/ja/server/database/sql/database.py
+++ b/src/ja/server/database/sql/database.py
@@ -1,5 +1,7 @@
+from copy import deepcopy
+from typing import List, Optional, Callable
 from datetime import datetime
-from typing import List
+from sqlalchemy import create_engine
 
 from ja.common.work_machine import ResourceAllocation
 from ja.common.docker_context import DockerContext, DockerConstraints, MountPoint
@@ -9,7 +11,7 @@ from ja.server.database.types.work_machine import WorkMachine, WorkMachineResour
     WorkMachineState
 from ja.server.database.database import ServerDatabase
 from sqlalchemy import Table, Column, Integer, String, MetaData, DateTime, Enum, ForeignKey, Boolean, ARRAY
-from sqlalchemy.orm import mapper, synonym, relationship
+from sqlalchemy.orm import mapper, synonym, relationship, sessionmaker, scoped_session, joinedload
 
 
 class SQLDatabase(ServerDatabase):
@@ -26,6 +28,9 @@ class SQLDatabase(ServerDatabase):
         @param user The username to use for the connection.
         @param password The password to use for the connection.
         """
+        self.scheduler_callback: Callable[["ServerDatabase"], None] = lambda *args: None
+        self.status_callback: Callable[["Job"], None] = lambda *args: None
+        self.in_scheduler_callback: bool = False
         if SQLDatabase._metadata:
             return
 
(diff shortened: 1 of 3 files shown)


ammen99:
I think tests for the callbacks need to be implemented. As well as checking that they are called when the database is updated, and that no infinite recursion happens.

nikolatzotchev:
> I think tests for the callbacks need to be implemented. As well as checking that they are called when the database is updated, and that no infinite recursion happens.

I'm not really sure how to test if there is no recursion happening.

ammen99:
> I'm not really sure how to test if there is no recursion happening.

You can do the following:

1. Set a callback which increments a counter each time it is called
2. Update the database to callback is triggered
3. In the callback, increment your counter and update the database again
4. If the callback is called and the counter is higher than 0, this means you're in a recursion

nikolatzotchev:
@ammen99 I fixed the bug here, it is not a problem right?

nikolatzotchev:
#101 btw

ammen99:
It's fine I think.

src/ja/server/database/sql/mock_database.py
```diff
@@ -24,93 +19,15 @@ def __init__(self) -> None:
...
-        if since is None:
-            return jobs
-        return [job for job in jobs if job.statistics.time_added >= since]
-
-    def set_scheduler_callback(self, callback: ServerDatabase.RescheduleCallback) -> None:
-        pass
-
-    def set_job_status_callback(self, callback: ServerDatabase.JobStatusCallback) -> None:
-        pass
+    def __del__(self) -> None:
```
ammen99: Is this needed? I thought Python will automatically call the parent's destructor.

src/ja/server/database/sql/mock_database.py
```diff
@@ -24,93 +19,15 @@ def __init__(self) -> None:
...
-    def get_jobs_on_machine(self, machine: WorkMachine) -> Optional[List[Job]]:
-        jobs: Optional[List[Job]] = self.session.query(Job).join(DatabaseJobEntry). \
-            join(WorkMachine, WorkMachine.uid == machine.uid).all()
-        return jobs
-
-    def assign_job_machine(self, job: Job, machine: WorkMachine) -> None:
-        job_entry = self.find_job_by_id(job.uid)
-        job_entry.assigned_machine = machine
-        self.session.commit()
+        def stat_call(j: Job) -> None:
```
ammen99: I think those things belong to the tests, not here.

src/test/server/database/test_database.py
```diff
@@ -49,38 +52,17 @@ def setUp(self) -> None:
                                         WorkMachineResources(ResourceAllocation(12, 32, 12)))
         self.work_machine2 = WorkMachine("machi2", WorkMachineState.ONLINE,
                                          WorkMachineResources(ResourceAllocation(2, 2, 2)))
+        self.work_machine3 = WorkMachine("machi1", WorkMachineState.ONLINE,
```
ammen99: Is the UID intentionally the same as machine1 ?

src/test/server/database/test_database.py
```diff
@@ -208,3 +196,40 @@ def test_query_jobs_false(self) -> None:
 
         jobs = self.mockDatabase.query_jobs(None, -1, self.work_machine)
         self.assertEqual(jobs, [self.mockDatabase.find_job_by_id(self.job.uid)])
+
+    def test_scheduler_callback(self) -> None:
+        call: Mock = Mock()
+        self.mockDatabase.set_scheduler_callback(call)
+        self.mockDatabase.update_job(self.job)
+        call.assert_called_with(self.mockDatabase)
```
ammen99: I think `assert_called_once_with` is even better.

ammen99: And in the other places as well.

src/ja/server/database/sql/database.py
```diff
@@ -149,30 +151,121 @@ def __init__(self, host: str = None, user: str = None, password: str = None):
             "statistics": synonym("_statistics", descriptor=DatabaseJobEntry.statistics),
             "machine": synonym("_machine", descriptor=DatabaseJobEntry.assigned_machine)
         })
+        if user is not None and password is not None and host is not None:
+            # example: "postgresql://pesho:pesho@127.0.0.1:5432/jobadd"
+            conn: str = "postgresql://" + user + ":" + password + "@" + host + "/jobadder"
+            self.engine = create_engine(conn)
```
ammen99: private attributes should begin with `_`

src/test/server/database/test_database.py
nikolatzotchev: yes, I just wanted to test when I update wm1 if it is going to break

src/ja/server/database/sql/mock_database.py
nikolatzotchev: I have no idea, but if that what are you telling me is true then it is not really needed.

PR #102, opened 2020-02-17
Implementation of Scheduler

### LINKED ISSUE #80: Scheduler
Implementation of the Scheduler class.
Needs #79 and #59

ammen99:
I found some bugs in the database, they are included in the last commit.
I think @nikolatzotchev will have to take a closer look and maybe submit the changes in another PR.

Fixes #80

### DIFF
--- a/src/ja/server/scheduler/scheduler.py
+++ b/src/ja/server/scheduler/scheduler.py
@@ -1,5 +1,10 @@
+from copy import deepcopy
+from ja.common.job import JobStatus
 from ja.server.database.database import ServerDatabase
-from ja.server.scheduler.algorithm import SchedulingAlgorithm
+from ja.server.database.types.work_machine import WorkMachineState
+from ja.server.dispatcher.dispatcher import Dispatcher
+from ja.server.scheduler.algorithm import SchedulingAlgorithm, get_allocation_for_job
+from typing import Dict
 
 
 class Scheduler:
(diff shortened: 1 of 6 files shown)


ammen99:
I realized that we also need to somehow call the dispatcher, so I just added it as a parameter to the scheduler.

Alternatively, the main class could fill this gap: we can make the Scheduler return its distribution, then the main class gets this and calls the dispatcher. However, I'd like to keep the main class as simple as possible, so I put it here.

src/ja/server/scheduler/scheduler.py
```diff
@@ -11,20 +16,64 @@ class Scheduler:
...
 
-        @param algorithm The scheduling algorithm to use to assign jobs to
-          work machines.
+        @param algorithm The scheduling algorithm to use to assign jobs to work machines.
+        @param dispatcher The dispatcher to use for sending commands to the work machines.
+        @param special_resources The available special resources.
         """
+        self._algorithm = algorithm
+        self._dispatcher = dispatcher
+        self._special_resources = deepcopy(special_resources)
```
nikolatzotchev: Why is there the need to use deepcopy.

src/ja/server/scheduler/scheduler.py
```diff
@@ -11,20 +16,64 @@ class Scheduler:
...
+        available_machines = list(filter(lambda m: m.state == WorkMachineState.ONLINE, database.get_work_machines()))
+        new_schedule = self._algorithm.reschedule_jobs(database.get_current_schedule(),
+                                                       available_machines, self.special_resources)
+
+        # Reset resources and recalculate them
+        for machine in available_machines:
+            machine.resources.deallocate(machine.resources.total_resources - machine.resources.free_resources)
+        self._special_resources = deepcopy(self._total_special_resources)
+
+        for job in new_schedule:
```
nikolatzotchev: isn't it going to be better to call this job -> job_entry, so that you don't have job.job.

src/ja/server/scheduler/scheduler.py
JohannesGaessler: Even if there is no specific need it is good practice I think.
Otherwise you might get strange bugs because two objects are changing the same list.

JohannesGaessler: On line 74 the special resources are being manipulated.
In this case you should definitely make a deepcopy I think.

src/ja/server/scheduler/scheduler.py
nikolatzotchev: this minor things, I will resolve it.

PR #104, opened 2020-02-17
server config

### LINKED ISSUE #70: ServerConfig
Configuration for the server.
Contains things like a list of special resources (different licenses and how many of them are available), and some options for the scheduling algorithm.

M1keReck:
mostly getters for the server config

fixes #70

### DIFF
--- a/src/ja/server/config.py
+++ b/src/ja/server/config.py
@@ -1,43 +1,74 @@
-from typing import Dict
-
+from typing import Dict, cast
 from ja.common.config import Config
 
 
-class DatabaseConfig(Config):
+class LoginConfig(Config):
     """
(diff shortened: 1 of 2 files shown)


src/ja/server/config.py
```diff
@@ -1,55 +1,126 @@
-from typing import Dict
-
+from typing import Dict, cast
 from ja.common.config import Config
 
 
-class DatabaseConfig(Config):
+# Database config and email config contain the same properties so they can be modelled by one class,
+# not sure about the name though.
+class AccessConfig(Config):
```
ammen99: Maybe LoginConfig would be a better name?

PR #105, opened 2020-02-17
New worker config

jooe1:


src/ja/worker/config.py
```diff
@@ -10,27 +9,73 @@ class WorkerConfig(Config):
     Config for the worker client.
     """
 
+    def __init__(self, uid: str, ssh_config: SSHConfig, resource_allocation: ResourceAllocation):
```
ammen99: The UID is not part of the config, it will be assigned by the server.

src/ja/worker/config.py
```diff
@@ -10,27 +9,73 @@ class WorkerConfig(Config):
...
+    def __eq__(self, o: object) -> bool:
+        if isinstance(o, WorkerConfig):
+            return self._ssh_config == o._ssh_config \
+                and self._uid == o._uid \
+                and self._resource_allocation == o._resource_allocation
+        else:
+            return False
+
     @property
     def uid(self) -> str:
```
ammen99: @JohannesGaessler why is there a UID in the worker? Isn't this assigned by the server? Or is this a label? I can imagine this being useful if you could specify a label like "room100_" and then UIDs are generated in the server as "room100_1, room100_2", etc.

src/test/worker/config.py
```diff
@@ -0,0 +1,80 @@
...
+        self.assertEqual(self._wmc.resources.cpu_threads, 1)
+        self.assertEqual(self._wmc.resources.memory, 2)
+        self.assertEqual(self._wmc.resources.swap, 3)
+
+    def test_from_dict(self) -> None:
+        d: Dict[str, object] = dict()
+        d["ssh_config"] = self._wmc.ssh_config.to_dict()
+        d["resource_allocation"] = self._wmc.resources.to_dict()
+        d["uid"] = self._wmc.uid
+        # self.assertDictEqual(self._wmc, d)
```
ammen99: Is the code commented out because the other PR is not merged yet?

src/ja/worker/config.py
jooe1: This was the same since the design phase, I m not sure why you mentioned it now since it was in the other PR. For me you are either trying to make me waste more time or you just trying to show that you are doing more reviews, either way I wont take your reviews seriously.

ammen99: 1) No design is perfect. I also had to change several things.

2) This particular change won't take you more than 2 minutes.

3) I am trying to point out inconsistencies, not the things you listed.

I am not sure why you show this attitude, I did not wish to offend you or waste your time.

src/ja/worker/config.py
JohannesGaessler: The worker needs to be able to set its own UID to be able to have a persistent UID upon program restart/system reboot.
The idea is that the worker will remember its UID so that it will register with the same UID the next time you connect to the server.
Otherwise the same worker registering to the server multiple times would have a different UID each time.
Or even worse, the workers would swap UIDs between each other.

src/ja/worker/config.py
JohannesGaessler: I did the design for the worker config.
I intentionally included a UID for the worker, see the thread below.

src/ja/worker/config.py
```diff
@@ -10,27 +9,73 @@ class WorkerConfig(Config):
     Config for the worker client.
     """
 
+    def __init__(self, uid: str, ssh_config: SSHConfig, resource_allocation: ResourceAllocation):
+        """!
+        creates resource allocation instance
+        """
+        if uid is None:
+            raise ValueError("missing uid")
```
JohannesGaessler: The worker UID should be allowed to be None.
In that case the UID should be assigned automatically by the server.

src/ja/worker/config.py
```diff
@@ -10,27 +9,73 @@ class WorkerConfig(Config):
...
+        if ssh_config is None:
+            raise ValueError("missing ssh config")
+        if resource_allocation is None:
+            raise ValueError("missing resources")
```
JohannesGaessler: As I've said before, please write more detailed error messages.
(Although I'm not sure you even need them in this case.)

src/ja/worker/config.py
```diff
@@ -10,27 +9,73 @@ class WorkerConfig(Config):
...
+        d: Dict[str, object] = dict()
+        d["ssh_config"] = self._ssh_config.to_dict()
+        d["resource_allocation"] = self._resource_allocation.to_dict()
+        d["uid"] = self._uid
+        return d
 
     @classmethod
     def from_dict(cls, property_dict: Dict[str, object]) -> "WorkerConfig":
-        pass
+        if "resource_allocation" in property_dict:
```
JohannesGaessler: `cls._get_dict_from_dict` automatically raises a ValueError if the dictionary does not contain the requested key.
There is no need for you to manually check if property_dict contains a key beforehand.

src/ja/worker/config.py
```diff
@@ -10,27 +9,73 @@ class WorkerConfig(Config):
...
+        else:
+            raise ValueError("missing ssh")
+        if "resource_allocation" in property_dict:
+            resource_allocation = ResourceAllocation.from_dict(
+                cls._get_dict_from_dict(property_dict=property_dict, key="resource_allocation", mandatory=False))
+        else:
+            raise ValueError("missing resource allocation")
+
+        if "uid" in property_dict:
+            uid = cls._get_str_from_dict(property_dict, "uid")
```
JohannesGaessler: The worker UID should not be a mandatory property.

src/ja/worker/config.py
```diff
@@ -13,12 +13,10 @@ def __init__(self, uid: str, ssh_config: SSHConfig, resource_allocation: Resourc
         """!
         creates resource allocation instance
         """
-        if uid is None:
-            raise ValueError("missing uid")
         if ssh_config is None:
-            raise ValueError("missing ssh config")
+            raise ValueError("the WorkerConfig class is missing the ssh_congfig param and it cannot be None!")
```
JohannesGaessler: ssh_congfig -> ssh_config

src/ja/worker/config.py
```diff
@@ -62,20 +60,10 @@ def to_dict(self) -> Dict[str, object]:
...
-
-        if "uid" in property_dict:
-            uid = cls._get_str_from_dict(property_dict, "uid")
-        else:
-            raise ValueError("missing uid")
+        ssh_config = SSHConfig.from_dict(
+            cls._get_dict_from_dict(property_dict=property_dict, key="ssh_config", mandatory=True))
+        resource_allocation = ResourceAllocation.from_dict(
+            cls._get_dict_from_dict(property_dict=property_dict, key="resource_allocation", mandatory=True))
+        uid = cls._get_str_from_dict(property_dict, "uid")
```
JohannesGaessler: The uid property is still mandatory because the default value for mandatory is True.
You need to specify mandatory=False.

src/ja/worker/config.py
```diff
@@ -64,6 +64,9 @@ def from_dict(cls, property_dict: Dict[str, object]) -> "WorkerConfig":
             cls._get_dict_from_dict(property_dict=property_dict, key="ssh_config", mandatory=True))
         resource_allocation = ResourceAllocation.from_dict(
             cls._get_dict_from_dict(property_dict=property_dict, key="resource_allocation", mandatory=True))
-        uid = cls._get_str_from_dict(property_dict, "uid", mandatory=False)
+        if "uid" in property_dict:
```
JohannesGaessler: Remove this check and instead add "uid" to the list of optional properties in the test that inherits from AbstractSerializableTest.

JohannesGaessler:
This PR is now looking good to merge.

Last thing to do:
Nikola and I had both implemented SSHConfig.
This PR is based on Nikolas version.
However, the master branch is using my version.
For you the difference is that the default value of key_filename has changed: "~/.ssh/id_rsa" -> None.
You need to change the test case that inherits from AbstractSerializableTest: give the SSHConfig object and self._object_dict the same value for key_filename.

ammen99:
Please be more careful when merging @JohannesGaessler, this PR should have been squashed.

JohannesGaessler:
Yes.
I really am sorry.

PR #108, opened 2020-02-19
cli and tests for it

### LINKED ISSUE #67: IServerProxy

### LINKED ISSUE #68: CLI Handler

### LINKED ISSUE #69: User -> Server messages
Messages which the UserClient sends to the Server.
They operate on the ServerDatabase.

from `ja.user.message`:

- [ ] add

- [ ] cancel

- [ ] query

### LINKED ISSUE #71: User Server Proxy
The real proxy which is used by the user to communicate to the server.

M1keReck:
New PR for the CLI (since I can't rebase on top of master in the other one)
All tests passed locally and you can actually try out the CLI by calling
`print(UserClientCLIHandler(path to yaml file).get_command_from_cli().to_dict())`

Now also includes user commands.

Fixes #71 
Fixes #69
Fixes #68 
Fixes #67


src/ja/user/cli.py
```diff
@@ -1,19 +1,210 @@
...
+        if args.command == "add":
+            # creating job instance
+            constraints: JobSchedulingConstraints = JobSchedulingConstraints(UserClientCLIHandler._pr[args.priority],
+                                                                             args.preemptible,
+                                                                             args.special_resources)
+            docker_constraints: DockerConstraints = DockerConstraints(args.threads, args.memory)
+            mount_points: List[MountPoint] = []
+            for i in range(0, len(args.mount), 2):
+                mount_points.append(MountPoint(args.mount[i], args.mount[i + 1]))
+            docker_context: DockerContext = DockerContext(args.source, mount_points)
```
JohannesGaessler: You are giving the source file **path** to the DockerContext constructor.
However, the DockerContext constructor expects to be given the source file **contents**.
You need to read from the source file and pass the contents on to the DockerContext constructor.

src/test/proxy/server_proxy.py
```diff
@@ -0,0 +1,142 @@
...
+        return ServerResponse(result_string="Successfully added job with id: %s" % job.uid,
+                              is_success=True)
+
+    def cancel_job(self, cancel_command: CancelCommand) -> ServerResponse:
+        job_uid: str = cancel_command.uid
+        job_label: str = cancel_command.label
+        if job_uid is not None:
+            for job in self._jobs:
+                if job.uid == job_uid:
+                    if False:  # TODO: Find a way to verify if this is the owner
```
JohannesGaessler: I think it's fine if you don't implement owner verification for the dummy.

src/ja/user/message/query.py
```diff
@@ -52,7 +52,7 @@ def label(self) -> List[str]:
         return self._label
 
     @property
-    def owner(self) -> List[str]:
+    def owner(self) -> List[int]:
```
JohannesGaessler: The owners should be a list of strings, not a list of ints.
This is the thing that the user types in (usernames).
At some point on the server these usernames can be translated into Linux user IDs.

src/ja/user/cli.py
```diff
@@ -1,19 +1,210 @@
...
+                                help="Assign a number of cpu threads required for your job.")
+        parser_add.add_argument("--memory", "-m",
+                                type=int, default=1,
+                                help="Assign a specific amount of memory to your job")
+        parser_add.add_argument("--preemptible", "--pe", action="store_true")
+        parser_add.add_argument("--blocking", "--bl", action="store_true")
+        parser_add.add_argument("--mount", nargs="+", type=str, default=[])
+        parser_add.add_argument("--email", "-e", required=email is None, default=email)
+        parser_add.add_argument("--special-resources", "--sr", nargs="+", default=[])
+        parser_add.add_argument("--owner", default=os.getuid(), type=int, help="Not to be used outside of testing!")
```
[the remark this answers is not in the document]
M1keReck: I am using this command to add jobs with different owners in the tests, so I can test querying by owners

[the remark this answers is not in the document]
JohannesGaessler: I see.
The owner ID should be determined by the username specified in the SSHConfig object after the job has been transferred to the server.
When you are finished with this PR, please remove this argument and just give the Job some value like -1 for the owner ID.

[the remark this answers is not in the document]
M1keReck: Couldn't I keep it? One test won't run without this option. It won't appear in the CLI since its help is suppressed, and if the user somehow finds a way to run it, it'll be overridden on the server.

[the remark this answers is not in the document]
JohannesGaessler: It's not the cleanest solution but considering you're already behind on schedule I think it's fine for you to do it like that.

M1keReck:
All tests pass locally, the travis checks throw the same error for all tests. Apparently the problem is the uid setter of the job, and the error is raised in the sql package for some reason.

`Traceback (most recent call last):
  File "/home/travis/build/DistributedTaskScheduling/JobAdder/src/test/user/cli.py", line 38, in setUp
    job.uid = str(i)
  File "<string>", line 1, in __set__
  File "/home/travis/build/DistributedTaskScheduling/JobAdder/src/ja/common/job.py", line 154, in uid
    self._uid = value
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/sqlalchemy/orm/attributes.py", line 268, in __set__
    instance_state(instance), instance_dict(instance), value, None
AttributeError: 'Job' object has no attribute '_sa_instance_state'`

src/ja/user/cli.py
M1keReck: Now the whole file is read and passed on to the constructor as a string, hope that's right

JohannesGaessler:
It seems to me like the unittests for the mock database are making some sort of permanent change to the Job class.
This is the responsibility of @nikolatzotchev to debug.

JohannesGaessler:
Disregard my last post, I found out what the problem is.
The problem is that Mike is setting commands and jobs as static properties of CLITest.
As such they get initialized on import, before the tests for the mock database are being run.
Presumably the mock database tests then do something weird.
If you just initialize commands and jobs as an attribute of the object in setUp (something like self.commands = ...) it works.
(The paths for p.yaml and add.yaml is broken when running the tests from the main test directory though.)

src/test/proxy/server_proxy.py
```diff
@@ -0,0 +1,138 @@
...
+        owner: List[str] = query_command.owner
+        priority: List[JobPriority] = query_command.priority
+        status: List[JobStatus] = query_command.status
+        is_preemptible: bool = query_command.is_preemptible
+        special_resources: List[List[str]] = query_command.special_resources
+        cpu_threads: Tuple[int, int] = query_command.cpu_threads
+        memory: Tuple[int, int] = query_command.memory
+        # before: datetime = query_command.before
+        # after: datetime = query_command.after
+
```
M1keReck: This will not be tested since the Jobs don't have an added property. I could however add it in the dummy somehow if that's really necessary.

ammen99:
One general issue that I noticed when compiling statistics about our project: you are missing unit tests for the Cancel command (also Query seems to be missing some lines, but maybe we can test them in the QA). Take a look here: https://travis-ci.org/DistributedTaskScheduling/JobAdder/jobs/657678748#L730

src/test/proxy/server_proxy.py
JohannesGaessler: I think it's fine if you don't test for the time

src/ja/user/message/add.py
```diff
@@ -13,13 +15,33 @@ def __init__(self, config: AddCommandConfig):
...
+    def execute(self, database: ServerDatabase) -> Response:
+        job: Job = self.config.job
+        db_job_id: DatabaseJobEntry = database.find_job_by_id(job.uid)
+        if db_job_id is not None:
+            return Response(result_string="Job with id %s already exists" % job.uid,
+                            is_success=False)
+        db_job_label: Job = database.find_job_by_label(job.label)
+        if db_job_label is not None:
+            return Response(result_string="Job with label %s already exists" % job.label,
+                            is_success=False)
```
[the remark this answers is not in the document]
M1keReck: I thought it may be necessary for when jobs are cancelled, unless we wanna cancel all jobs with a specific label or júst pick the first one.

JohannesGaessler:
The branch you force-pushed is based off of an old commit of the master branch.
Did you perhaps forget to pull from the master branch before rebasing?

src/ja/user/cli.py
```diff
@@ -1,19 +1,244 @@
...
+        parser_add = subparsers.add_parser("add")
+        parser_add.add_argument("--config", "-c", help="Config file path for adding a job.")
+        parser_add.add_argument("--label", "-l", help="Label for the job.")
+        parser_add.add_argument("--source", "-s", "--path", help="Path of the source file containing the job.")
+        parser_add.add_argument("--priority", "-p", choices="low medium high urgent 0 1 2 3".split())
+        parser_add.add_argument("--threads", "-t", type=int, help="Assign a number of cpu threads for your job.")
+        parser_add.add_argument("--memory", "-m", type=int, help="Assign a specific amount of memory to your job")
+        parser_add.add_argument("--preemptible", "--pe", action="store_true")
+        parser_add.add_argument("--blocking", "--bl", action="store_true")
+        parser_add.add_argument("--mount", nargs="+", type=str)
```
ammen99: I think it would be nice to add `action='append'`, this way the user can run `--mount A B --mount C D` which makes it more clear (also, this way you can just specify `nargs='2'` to make sure you always get pairs)

src/ja/user/cli.py
```diff
@@ -1,19 +1,244 @@
...
+                                  help="Job priorities to filter query results by.")
+        parser_query.add_argument("--status",
+                                  nargs="+",
+                                  choices="new queued running paused done killed crashed 0 1 2 3 4 5 6".split(),
+                                  help="Job status(es) to filter query results by.")
+        parser_query.add_argument("--preemptible", "--is-preemptible", "--ip",
+                                  dest="preemptible",
+                                  choices="false true t f".split(),
+                                  help="Get job(s) that can be preempted by other jobs.")
+        parser_query.add_argument("--special-resources", "--sr", help="Filter by the special resources of a job. \
```
ammen99: What do the lists separated by '.' represent? Should be clarified by the help message.

src/ja/user/message/add.py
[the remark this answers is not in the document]
ammen99: I also think labels don't have to be unique, but the cancel command could fail if multiple jobs with the same label are currently running. (Or maybe we can say that killing by label kills all such jobs?)

src/ja/user/message/cancel.py
```diff
@@ -59,5 +68,30 @@ def from_dict(cls, property_dict: Dict[str, object]) -> "CancelCommand":
...
+                return Response(result_string="Job with uid %s does not exist!" % self.uid,
+                                is_success=False)
+            else:
+                job = job_entry.job
+        if self.label is not None:
+            job = database.find_job_by_label(self.label)
+            if job is None:
+                return Response(result_string="Job with label %s does not exist!" % self.label,
+                                is_success=False)
+        job.status = JobStatus.CANCELLED
```
ammen99: What happens if the job is already finished? Or has been cancelled previously? That should be an error as well.

src/ja/user/message/cancel.py
```diff
@@ -59,5 +68,30 @@ def from_dict(cls, property_dict: Dict[str, object]) -> "CancelCommand":
...
+        database.update_job(job)
+        if self.uid is not None:
+            return Response(result_string="Successfully removed job with uid: %s" % self.uid,
```
ammen99: removed -> cancelled

src/ja/user/message/cancel.py
```diff
@@ -59,5 +68,30 @@ def from_dict(cls, property_dict: Dict[str, object]) -> "CancelCommand":
...
+                            is_success=True)
+        if self.label is not None:
+            return Response(result_string="Successfully removed job with label: %s" % self.label,
```
ammen99: removed -> cancelled

src/ja/user/message/query.py
```diff
@@ -259,5 +260,42 @@ def from_dict(cls, property_dict: Dict[str, object]) -> "QueryCommand":
                          special_resources_list, cpu_threads, memory, before, after)
         return a
 
-    def execute(self, database: ServerDatabase) -> "ServerResponse":
-        pass
+    def execute(self, database: ServerDatabase) -> Response:
+        jobs: List[DatabaseJobEntry] = database.query_jobs(None, -1, None)  # Query all DatabaseJobEntries
```
ammen99: I think you should try to pass as many arguments as possible to the query_jobs command, since SQL is faster than ourselves :)

src/test/proxy/server_proxy.py
```diff
@@ -0,0 +1,128 @@
...
+        job.status = JobStatus.QUEUED
+        return Response(result_string="Successfully added job with id: %s" % job.uid,
+                        is_success=True, uid=job.uid)
+
+    def cancel_job(self, cancel_command: CancelCommand) -> Response:
+        job_uid: str = cancel_command.uid
+        job_label: str = cancel_command.label
+        if job_uid is not None:
+            for job in self._jobs:
+                if job.uid == job_uid:
```
ammen99: Same thing as in the other file. If the job status is already cancelled, I do not think this operation should succeed.

src/ja/user/cli.py
```diff
@@ -1,19 +1,244 @@
...
+            label = UserClientCLIHandler._get_option("label", label, None, add_file)
+            source = UserClientCLIHandler._get_option("source", source, None, add_file)
+            priority = UserClientCLIHandler._get_option("priority", priority, "1", add_file)
+            threads = UserClientCLIHandler._get_option("threads", threads, -1, add_file)
+            memory = UserClientCLIHandler._get_option("memory", memory, 1, add_file)
+            mount = UserClientCLIHandler._get_option("mount", mount, [], add_file)
+            special_resources = UserClientCLIHandler._get_option("special-resources",
+                                                                 special_resources, [], add_file)
+            if "blocking" in add_file:
+                blocking = blocking or add_file["blocking"]
```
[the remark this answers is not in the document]
M1keReck: `"--preemptible", "--pe", action="store_true"`
Note that these options are never None, so the option in the config file would always be overridden. If the file says "true", the default value in the args "false" should be overridden.

src/ja/user/cli.py
```diff
@@ -1,19 +1,244 @@
...
+        # parser for add command.
+        parser_add = subparsers.add_parser("add")
+        parser_add.add_argument("--config", "-c", help="Config file path for adding a job.")
+        parser_add.add_argument("--label", "-l", help="Label for the job.")
+        parser_add.add_argument("--source", "-s", "--path", help="Path of the source file containing the job.")
+        parser_add.add_argument("--priority", "-p", choices="low medium high urgent 0 1 2 3".split())
+        parser_add.add_argument("--threads", "-t", type=int, help="Assign a number of cpu threads for your job.")
+        parser_add.add_argument("--memory", "-m", type=int, help="Assign a specific amount of memory to your job")
+        parser_add.add_argument("--preemptible", "--pe", action="store_true")
+        parser_add.add_argument("--blocking", "--bl", action="store_true")
```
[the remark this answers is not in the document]
M1keReck: I won't do that yet because I'd have to change some stuff in the tests as well

src/ja/server/database/database.py
```diff
@@ -20,6 +20,15 @@ def find_job_by_id(self, job_id: str) -> DatabaseJobEntry:
         @return The stored data about the job, or None if no such job was found.
         """
 
+    @abstractmethod
+    def find_job_by_label(self, label: str) -> Job:
```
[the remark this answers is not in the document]
nikolatzotchev: Ok but I have to change the database so that this field is not unique.

src/ja/user/cli.py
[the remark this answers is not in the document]
M1keReck: NOT preemptible!

src/ja/server/database/database.py
```diff
@@ -20,6 +20,15 @@ def find_job_by_id(self, job_id: str) -> DatabaseJobEntry:
         @return The stored data about the job, or None if no such job was found.
         """
 
+    @abstractmethod
+    def find_job_by_label(self, label: str) -> List[Job]:
```
ammen99: One last thing, I believe we should also have a unit test in the Database tests for this function as well.

JohannesGaessler:
After a squash and a rebase I think this is good to merge.

ammen99:
@M1keReck You should also rebase on top of the master branch to fix most of the errors, and I also think you have one line which is too long, after that we can merge.

M1keReck:
I actually wanted to add those but I am struggling with the MockDatabase initialization, if I can't solve it tonight, I'll bring this up in the meeting.

PR #109, opened 2020-02-22
Implementation ssh

JohannesGaessler:
I accidentally merged https://github.com/DistributedTaskScheduling/JobAdder/pull/93 when the travis build was failing.
The changes were reverted and this PR will be the same content but with working builds.


ammen99:
Seems like the WIP commit should be squashed with something else, didn't notice that one before :)

PR #112, opened 2020-02-28
New worker main

### LINKED ISSUE #65: WorkerMain
The main class for the worker, using a mock server proxy from #64 

Should also listen on the local socket for new commands from the server (that part could probably be shared with the server, should be investigated)

jooe1:
Fixes #65


src/ja/common/message/worker_commands/cancel_job.py
```diff
@@ -9,17 +9,27 @@
...
+        """
+        self._job_uid = job_uid
+
     def execute(self, worker_client: JobWorker) -> WorkerResponse:
         """!
         Cancel the job on the worker machine using the provided @worker_client
         @param worker_client:  the Worker object to use for the execution
         @return: a WorkerResponse with the appropriate response
         """
+        return worker_client.docker.cancel_job(self._job_uid)
```
JohannesGaessler: The DockerInterface does **not** create Response objects; its methods do not return anything.
The execute method needs to create these Response objects itself.

src/ja/common/message/worker_commands/pause_job.py
```diff
@@ -9,17 +9,27 @@
...
+        """
+        self._job_uid = job_uid
+
     def execute(self, worker_client: JobWorker) -> WorkerResponse:
         """!
         Pause the job on the worker machine using the provided @worker_client
         @param worker_client:  the Worker object to use for the execution
         @return: a WorkerResponse with the appropriate response
         """
+        return worker_client.docker.pause_job(self._job_uid)
```
JohannesGaessler: Same as CancelJobCommand.

src/ja/common/message/worker_commands/start_job.py
```diff
@@ -5,21 +5,32 @@
...
+        """
+        self._job = job
+
     def execute(self, worker_client: JobWorker) -> WorkerResponse:
         """!
         Start the job on the worker machine using the provided @worker_client
         @param worker_client:  the Worker object to use for the execution
         @return: a WorkerResponse with the appropriate response
         """
+        return worker_client.docker.add_job(self._job)
```
JohannesGaessler: Same as CancelJobCommand.

src/ja/worker/proxy/message_handler.py
```diff
@@ -13,7 +22,27 @@ class WorkerCommandHandler(CommandHandler):
...
+        "GetWorkloadCommand": GetWorkloadCommand,
+        "PauseJobCommand": PauseJobCommand,
+        "ChangeFiltersCommand": ChangeFiltersCommand,
+        "StartJobCommand": StartJobCommand
+    }
+
+    def _execute_command(self, command: WorkerCommand) -> Dict[str, object]:
+        return command.execute(self._worker_Client).to_dict()
+
+    def _process_message(self, command_dict: Dict[str, object], type_name: str, user: str) -> Dict[str, object]:
```
JohannesGaessler: _process_message (or main_loop as it was renamed by Ilia) needs to verify user permissions.
For the worker client this means checking if the user is part of the admin group (as was done in the server main class).

src/ja/common/message/worker_commands/cancel_job.py
jooe1: I was planning on using the docker.errors.APIError to ckeck if the command was successful or not. Bc the execute method can't be sure.

src/ja/worker/proxy/message_handler.py
jooe1: I m not sure why the checking of the user is happening on the worker when it can be detected while handling his command on the server.

JohannesGaessler: You have to check permissions on the worker because otherwise any user logged into the worker could directly call the remote and send any sort of command to the worker even if such a user is not an administrator.
Checking permissions on the worker has nothing to do with the human user specified in the job.
You only need to check whether the user sending a command to the worker is part of the admin group (the server will use an admin account to log in to the worker and send the command).

src/ja/common/message/worker_commands/cancel_job.py
JohannesGaessler: That sounds like a good solution.

src/ja/worker/main.py
```diff
@@ -2,16 +2,40 @@
...
+        self._worker_server_proxy = WorkerServerProxy(self._ssh_config)
+        self._docker = DockerInterface(self._worker_server_proxy)
+
+    @property
+    def docker(self) -> DockerInterface:
+        return self._docker
+
+    def run(self) -> None:
+        """!
+        Run the main loop of the server daemon.
```
ammen99: server -> worker

src/ja/worker/proxy/message_handler.py
```diff
@@ -13,7 +22,27 @@ class WorkerCommandHandler(CommandHandler):
...
-    def __init__(self) -> None:
+    def __init__(self, worker_client: JobWorker) -> None:
         """!
         Creates the socket and listens on it.
         """
+        self._worker_Client = worker_client
+
+    _server_commands = {
+        "CancelJobCommand": CancelJobCommand,
+        "GetWorkloadCommand": GetWorkloadCommand,
```
ammen99: you can remove get workload for now, since we won't implement it

src/ja/worker/proxy/message_handler.py
ammen99: You can see the `user_is_admin` function from the server. I think you can move it to the base CommandHandler so that we don't duplicate it.

M1keReck:
This could be closed right?

ammen99:
> This could be closed right?

Yes, Johannes takes this task over.

PR #116, opened 2020-03-06
Basic logging impl on the server module.

nikolatzotchev:
Started implementing the logger on the server, using the python [logger](https://docs.python.org/3/library/logging.html#module-logging), this provides also levels, for now I'm using only the INFO one. The log is send to a file, is this enough or should it else be send to stdou or something else. Will add the same stuff the the worker and user later this week.

### DIFF
--- a/src/ja/server/database/sql/database.py
+++ b/src/ja/server/database/sql/database.py
@@ -13,6 +13,10 @@ from sqlalchemy import Table, Column, Integer, String, MetaData, DateTime, Enum,
 from sqlalchemy.orm import mapper, synonym, relationship, sessionmaker, scoped_session, joinedload
 from ja.common.proxy.ssh import SSHConfig
 
+import logging
+
+logger = logging.getLogger(__name__)
+
 
 class SQLDatabase(ServerDatabase):
     """!
@@ -160,6 +164,7 @@ class SQLDatabase(ServerDatabase):
             self.engine = create_engine(_conn)
             self.scoped = scoped_session(sessionmaker(self.engine))
             SQLDatabase._metadata.create_all(self.engine)
+            logger.info("connection on %s to the database: %s" % (host, database_name))
 
     def __del__(self) -> None:
         self.scoped.remove()  # type: ignore
@@ -177,6 +182,10 @@ class SQLDatabase(ServerDatabase):
                 jobs_entry.statistics.paused_time \
                     = (datetime.now() - jobs_entry.statistics.time_started).seconds - jobs_entry.statistics.running_time
             session.commit()
+            logger.info("job entry with job id: %s found." % job_id)
+            logger.debug(str(jobs_entry.job))
+        else:
+            logger.error("job with id: %s not found" % job_id)
         return jobs_entry
 
     def find_job_by_label(self, label: str) -> List[Job]:
@@ -196,6 +205,8 @@ class SQLDatabase(ServerDatabase):
                                                                     0, 0),
                                          machine=None)
             session.add(job_entry)
+            logger.info("first add for job: %s" % job.uid)
+            logger.debug(str(job))
         else:
             if old_job.status == JobStatus.PAUSED and job.status != JobStatus.PAUSED:
                 old_job_entry.statistics.paused_time = \
@@ -212,6 +223,8 @@ class SQLDatabase(ServerDatabase):
                 if job.status != old_job.status:
                     old_job_entry.job.status = job.status
                     self.status_callback(job)
+            logger.info("update job: %s" % job.uid)
+            logger.debug("old job: \n%s \n new job: \n%s" % (str(old_job), str(job)))
         session.commit()
         self._call_scheduler()
 
@@ -219,6 +232,10 @@ class SQLDatabase(ServerDatabase):
         session = self.scoped()
         jobs: Optional[List[Job]] = session.query(Job).join(DatabaseJobEntry). \
             join(WorkMachine, WorkMachine.uid == machine.uid).options(joinedload("*")).all()
+        if jobs is not None:
+            logger.info("jobs on machine %s: " % machine.uid + str([job.uid for job in jobs]))
+        else:
+            logger.info("no jobs on machine %s" % machine.uid)
         return jobs
 
     def assign_job_machine(self, job: Job, machine: WorkMachine) -> None:
@@ -226,6 +243,7 @@ class SQLDatabase(ServerDatabase):
         job_entry = self.find_job_by_id(job.uid)
         job_entry.assigned_machine = machine
         session.commit()
+        logger.info("assign job: %s, to machine %s" % (job.uid, (machine.uid if machine is not None else None)))
         self._call_scheduler()
 
     def update_work_machine(self, machine: WorkMachine) -> None:
@@ -233,7 +251,12 @@ class SQLDatabase(ServerDatabase):
         work_machine: WorkMachine = session.query(WorkMachine).filter(WorkMachine.uid == machine.uid).first()
         if work_machine is None:
             session.add(machine)
+            logger.info("adding work machine: %s" % machine.uid)
+            logger.debug(str(machine))
         else:
+            logger.info("updated work machine with uid: %s" % machine.uid)
+            logger.debug(
+                "old machine: \n %s \n new machine: \n %s" % (str(machine), str(work_machine)))
             work_machine.state = machine.state
             work_machine.ssh_config = machine.ssh_config
             work_machine.resources = machine.resources
@@ -261,6 +284,8 @@ class SQLDatabase(ServerDatabase):
         if user_id != -1:
             jobs_query = jobs_query.join(Job).filter(Job.owner_id == user_id)
         jobs: List[DatabaseJobEntry] = jobs_query.all()
+        if len(jobs) == 0:
+            logger.info("no jobs found")
         if since is None:
             return jobs
         return [job for job in jobs if job.statistics.time_added >= since]
@@ -268,6 +293,7 @@ class SQLDatabase(ServerDatabase):
     def _call_scheduler(self) -> None:
         if not self.in_scheduler_callback:
             self.in_scheduler_callback = True
+            logger.info("calling scheduler callback")
             self.scheduler_callback(self)
             self.in_scheduler_callback = False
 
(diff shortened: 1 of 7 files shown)


nikolatzotchev:
fuck, ok they are some test not passing, that is because @ammen99 doesn't fill whole jobs when using them in the tests, should we fix that, or maybe disable logging in some cases.

src/ja/server/database/sql/database.py
```diff
@@ -177,6 +182,10 @@ def find_job_by_id(self, job_id: str) -> Optional[DatabaseJobEntry]:
                 jobs_entry.statistics.paused_time \
                     = (datetime.now() - jobs_entry.statistics.time_started).seconds - jobs_entry.statistics.running_time
             session.commit()
+            logger.info("job entry with job id: %s found." % job_id)
+            logger.debug("%s" % str(jobs_entry.job.to_dict()))
```
JohannesGaessler: Couldn't you just use str(jobs_entry.job) since Job has implemented __str__() ?

src/ja/server/proxy/proxy.py
```diff
@@ -70,22 +73,34 @@ def uid(self) -> str:
 
     def dispatch_job(self, job: Job) -> WorkerResponse:
         command = StartJobCommand(job)
+        logger.info("dispatching job")
```
JohannesGaessler: Please attach job uids to the info level.

src/ja/server/database/sql/database.py
nikolatzotchev: yea probably, sorry for the late replay I broke my linux groups on my laptop, and after some trying I reinstalled, at least all the tests are passing locally now.

nikolatzotchev:
I don't get it, how all of a sudden there are 20 mypy errors when I didn't change anything involving them.

nikolatzotchev:
2020-03-11-132713_grim
Sometimes I'm really confused.

nikolatzotchev:
I will add logging for the other components in other pr.

nikolatzotchev:
I never logged errors, but I don't see many useful error cases to log on the server. If you have any idea, let me know.

src/ja/server/proxy/command_handler.py
```diff
@@ -60,9 +63,12 @@ def _user_is_admin(self, user: str) -> bool:
         return self._admin_group in groups
 
     def _execute_command(self, command: ServerCommand) -> Dict[str, object]:
+        logger.info("executing server command")
+        logger.debug("%s" % str(command))
         return command.execute(self._database).to_dict()
```
M1keReck: Wouldn't it be good to log the output of the command execution like you you did in `WokerProxy`? Or maybe specify if it succeeded or not.

nikolatzotchev: Maybe, but I tried to add the logging stuff without formatting the existed code. But maybe I can change it a bit.

nikolatzotchev: I addressed this in my latest fix.

src/ja/server/database/sql/database.py
```diff
@@ -160,6 +164,7 @@ def __init__(self, host: str = None, user: str = None, password: str = None):
             self.engine = create_engine(_conn)
             self.scoped = scoped_session(sessionmaker(self.engine))
             SQLDatabase._metadata.create_all(self.engine)
+            logger.info("connection to real database")
```
ammen99: Maybe print the name of the database and the IP? (username and password are confidential so not a good fit).

src/ja/server/database/sql/database.py
```diff
@@ -177,6 +182,10 @@ def find_job_by_id(self, job_id: str) -> Optional[DatabaseJobEntry]:
                 jobs_entry.statistics.paused_time \
                     = (datetime.now() - jobs_entry.statistics.time_started).seconds - jobs_entry.statistics.running_time
             session.commit()
+            logger.info("job entry with job id: %s found." % job_id)
+            logger.debug("%s" % str(jobs_entry.job))
```
ammen99: just `logger.debug(str(jobs_entry.job))` ?

src/ja/server/database/sql/database.py
```diff
@@ -177,6 +182,10 @@ def find_job_by_id(self, job_id: str) -> Optional[DatabaseJobEntry]:
...
+        else:
+            logger.info("job with id: %s not found" % job_id)
```
ammen99: I think it would be better to `logger.error` here...

src/ja/server/database/sql/database.py
```diff
@@ -196,6 +205,8 @@ def update_job(self, job: Job) -> None:
                                                                     0, 0),
                                          machine=None)
             session.add(job_entry)
+            logger.info("first add for job: %s" % job.uid)
+            logger.debug("%s" % str(job))
```
ammen99: Same as above, what is the point of not directly printing str(job) ?

src/ja/server/database/sql/database.py
```diff
@@ -212,28 +223,40 @@ def update_job(self, job: Job) -> None:
...
+        logger.info("assign job: %s, to machine %s" % (job.uid, (machine.uid if machine is not None else None)))
         self._call_scheduler()
 
     def update_work_machine(self, machine: WorkMachine) -> None:
         session = self.scoped()
         work_machine: WorkMachine = session.query(WorkMachine).filter(WorkMachine.uid == machine.uid).first()
         if work_machine is None:
             session.add(machine)
+            logger.info("adding work machine: %s" % machine.uid)
+            logger.debug("%s" % str(machine))
```
ammen99: just str(machine)

src/ja/server/main.py
```diff
@@ -30,6 +33,7 @@ def _init_algorithm() -> SchedulingAlgorithm:
     @staticmethod
     def _read_config() -> ServerConfig:
         with open('/etc/jobadder/server.conf') as f:
+            logger.info("read config")
```
ammen99: I think a more helpful message would include the name of the config file being read.

src/ja/server/proxy/command_handler.py
```diff
@@ -60,9 +64,14 @@ def _user_is_admin(self, user: str) -> bool:
         return self._admin_group in groups
 
     def _execute_command(self, command: ServerCommand) -> Dict[str, object]:
-        return command.execute(self._database).to_dict()
+        logger.info("executing server command")
+        logger.debug("%s" % str(command))
```
ammen99: just str(..)

src/ja/server/proxy/command_handler.py
```diff
@@ -60,9 +64,14 @@ def _user_is_admin(self, user: str) -> bool:
...
+        r_dict = command.execute(self._database)
+        logger.debug("%s" % str(r_dict))
```
ammen99: same

ammen99: Also, you write something like `"Response: %s" % str(r_dict)` to be clear that you're printing the response.

src/ja/server/proxy/command_handler.py
```diff
@@ -60,9 +64,14 @@ def _user_is_admin(self, user: str) -> bool:
...
+        return r_dict.to_dict()
 
     def _process_worker_message(self, command_dict: Dict[str, object], type_name: str, user: str) -> Dict[str, object]:
+        logger.info("processing worker message")
```
ammen99: if you look at the logic below, worker_message is called always, and fails if the param is not a user message ...
So I do not think it makes sense to print something here, as well as below.

src/ja/server/database/sql/database.py
nikolatzotchev: but then if the level is set to INFO you wont get any logging.

src/ja/server/database/sql/database.py
nikolatzotchev: but the database will be always called jobadder or?

src/ja/server/main.py
```diff
@@ -30,6 +33,7 @@ def _init_algorithm() -> SchedulingAlgorithm:
     @staticmethod
     def _read_config() -> ServerConfig:
         with open('/etc/jobadder/server.conf') as f:
+            logger.info("reading server.conf config file")
```
ammen99: reading `/etc/jobadder/server.conf` config file ... You could have a thousand different server.confs on your computer, but only one of them is loaded!

src/ja/server/database/sql/database.py
ammen99: I guess this is something johannes should change in his PR. But otherwise, it is still good to see that the values are read from the config file correctly.

nikolatzotchev: yes he changed it in his pr, should I then wait for it to be merged before solving this one.

nikolatzotchev:
I moved the logger ini, so that it will be simpler to implement with the other components (user, workers, common) but I'm planning on doing them in different PR.

src/ja/server/proxy/command_handler.py
```diff
@@ -50,7 +55,11 @@ def __init__(self, database: ServerDatabase, admin_group: str):
     }
 
     def _execute_command(self, command: ServerCommand) -> Dict[str, object]:
-        return command.execute(self._database).to_dict()
+        logger.info("executing server command")
```
JohannesGaessler: Please add the name of the command class to the info level (something like `type(command).__name__`)

src/ja/server/proxy/command_handler.py
```diff
@@ -50,7 +55,11 @@ def __init__(self, database: ServerDatabase, admin_group: str):
...
+        logger.debug(str(command))
+        r_dict = command.execute(self._database)
```
JohannesGaessler: Please add whether the command was successful to the info level.

nikolatzotchev:
should be good now, what should be the default logging level, error or?

JohannesGaessler:
I'm not entirely sure.
I think we should set the default logging level to debug because the likelihood of our software containing bugs is still relatively high.

ammen99:
> I think we should set the default logging level to debug because the likelihood of our software containing bugs is still relatively high.

I'd say we should be able to set the level with an env var or something like that, and default to error. Consider that the log is stdout, so this will spit out messages in the syslog when started with systemd.

nikolatzotchev:
@JohannesGaessler I think I cant merge this because you have some requested changes, but I thing I fixed them already.

ammen99:
Well this has 2 approving reviews, and @JohannesGaessler's comments have been addressed, so let's move this forward.

PR #117, opened 2020-03-07
tests: use paths relative to the test itself

ammen99:
Relying on the current directory may break if we run tests from different directories.

I had a problem on my computer since I run the tests a little bit differently than the CI :)
Also, I think we need to specify which docker version we need. For example, I had docker 1.13 or something like that, and it wasn't working. @JohannesGaessler Do you know which minimum version we need?


nikolatzotchev:
I managed to run all test except the one that tests the mount point, I get no such file, for output_file.

nikolatzotchev:
I have docker 18.09.8

ammen99:
> I managed to run all test except the one that tests the mount point, I get no such file, for output_file.

Were the tests working without this PR?

JohannesGaessler:
I don't know the minimum Docker version required.
I have 19.03.6-ce installed locally and the tests were working.

PR #119, opened 2020-03-07
Server shutdown

### LINKED ISSUE #110: Figure out a way to close the server/worker daemons gracefully
- [ ] Setup & Start server

- [ ] Add stop command to the command handler & verify it works
 
- [ ] Systemd integration (service file, with sql deps and graceful stop)

- [ ] Installation of the above (maybe a script?)

ammen99:
Fixes #110 

- [x] Mark all jobs as crashed which are still running.


ammen99:
I have not been able to figure out how to make setup.py exclude the `test/` folder. I have tried multiple possible solutions (`exclude="test/"` and adding a `MANIFEST.in` with `prune/recursive-exclude`) but nothing seems to work. As a workaround I just rename the `__init__.py` file so that the folder does not get recognized as a source.

Also, I am using a separate script. I think this will be easier than trying to hack setup.py (somewhere in the docs I read setuptools is not supposed to install files to `/etc` which is needed for systemd service files and config files)

install.sh
```diff
@@ -0,0 +1,34 @@
...
+./setup.py install --prefix=$PREFIX
+mv ./test/init ./test/__init__.py
+cd $SCRIPTPATH
+
+# TODO: when remote is ready, set SUID
+
+# Install Systemd unit files
+SERVER_SRV_FILE=/etc/systemd/system/JobAdderServer.service
+
+cp ./data/bin/JobAdderServer.service $SERVER_SRV_FILE
```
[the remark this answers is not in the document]
ammen99: I do not think we need a symbolic link.

install.sh
```diff
@@ -0,0 +1,34 @@
...
+if [[ $EUID -ne 0 ]]; then
+   echo "This script must be run as root."
+   exit 1
+fi
+
+PREFIX=/usr
+SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+
+# Install JobAdder source code + binaries
+# FIXME: how to make setup.py ignore test folder??
```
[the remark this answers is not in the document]
ammen99: No, at least not in my tests. If you manage to get it working, I can add the changes to this PR, however I attempted this and also adding a special MANIFEST.in, nothing seemed to work.

install.sh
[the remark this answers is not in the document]
JohannesGaessler: I'm asking because the service files on my system are almost all symbolic links; I thought there may be a reason for that.

install.sh
```diff
@@ -0,0 +1,34 @@
...
+cd $SCRIPTPATH/src/
+mv ./test/__init__.py ./test/init
+./setup.py install --prefix=$PREFIX
```
JohannesGaessler: Shouldn't we be using pip to install our project?
I think setup.py install doesn't support automatic uninstallation.

data/config-files/server.conf
```diff
@@ -0,0 +1,16 @@
...
+        username: postgres
+        password: test123
+email_config:
+        host: 127.0.0.1
+        port: 1
+        username: ammen99
+        password: ammen99
+special_resources:
+        test: 10
+blocking_enabled: True
```
M1keReck: Do we still need this attribute?

install.sh
ammen99: I changed to pip, thanks for the suggestion.

install.sh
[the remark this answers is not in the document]
ammen99: The same thing happens when installing with pip3, so I am leaving this for now ...

data/config-files/server.conf
ammen99: yes, because it is required by the ServerConfig class here: https://github.com/DistributedTaskScheduling/JobAdder/blob/master/src/ja/server/config.py#L141

ammen99:
Seems like TravisCI is having some issues at the moment, however, builds are still being run (albeit late). This PR for example is passing: https://travis-ci.org/github/DistributedTaskScheduling/JobAdder/builds

src/ja/server/proxy/command_handler.py
```diff
@@ -57,11 +58,20 @@ def __init__(self, database: ServerDatabase, admin_group: str):
...
         groups = [g.gr_name for g in grp.getgrall() if user in g.gr_mem]
-        return self._admin_group in groups
+        # seems like the user is not reported to be in its own group ...
+        return self._admin_group in groups or user == self._admin_group
 
     def _execute_command(self, command: ServerCommand) -> Dict[str, object]:
         return command.execute(self._database).to_dict()
 
+    def _process_exit_message(self, type_name: str, user: str) -> Dict[str, object]:
+        if type_name != 'KillCommand':
```
nikolatzotchev: 'KillCommand'->"KillCommand", same thing for the other strings

ammen99: fixed.

install.sh
```diff
@@ -0,0 +1,37 @@
...
+   exit 1
+fi
+
+PREFIX=/usr
+SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+
+# Install JobAdder source code + binaries
+# FIXME: how to make setup.py ignore test folder??
+cd $SCRIPTPATH
+mv ./src/test/__init__.py ./src/test/init
```
JohannesGaessler: Does this prevent the installation of test?

install.sh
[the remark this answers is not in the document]
JohannesGaessler: Actually, now that I think about it we should definitely copy the service file instead of linking it.
Otherwise the service would be probably be linked to a file in the user space.

install.sh
ammen99: Yes, without `__init__.py` is is not detected as python module and not installed.

PR #120, opened 2020-03-08
Timeout for work machines and server

### LINKED ISSUE #106: Handle SSH connection timeout in WorkerProxy
If the connection to a work machine suddenly drops, we should eventually get a timeout on the SSH connection (or we may want to periodically send pings). Anyway, when this happens, it would make sense to mark the machine as offline and probably mark the jobs on  it as `CRASHED`.

Would be nice to have.

M1keReck:
Marks work machines as offline whenever the **Dispatcher** could not send it's commands in due time (120 secs) and informs the user if a connection to the server couldn't be established.

fixes #106

### DIFF
--- a/src/ja/server/dispatcher/dispatcher.py
+++ b/src/ja/server/dispatcher/dispatcher.py
@@ -1,8 +1,10 @@
 from ja.common.job import JobStatus
 from ja.server.database.database import ServerDatabase
 from ja.server.database.types.job_entry import DatabaseJobEntry
+from ja.server.database.types.work_machine import WorkMachine
 from ja.server.dispatcher.proxy_factory import WorkerProxyFactoryBase
-from typing import Dict
+from typing import Dict, List
+from paramiko.ssh_exception import SSHException  # type: ignore
 
 import logging
 logger = logging.getLogger(__name__)
@@ -22,6 +24,7 @@ class Dispatcher:
         """
         self._proxy_factory = proxy_factory
         self._previous_statuses: Dict[str, JobStatus] = dict()
+        self._lost_work_machines: List[WorkMachine] = []
 
     _job_status_order = [JobStatus.QUEUED, JobStatus.CANCELLED, JobStatus.PAUSED, JobStatus.RUNNING]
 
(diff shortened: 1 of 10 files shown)


src/ja/server/dispatcher/dispatcher.py
```diff
@@ -36,17 +38,35 @@ def set_distribution(self, job_distribution: JobDistribution) -> None:
...
                     new_statuses[job.uid] = job.status
                 elif job.status == JobStatus.QUEUED:
                     new_statuses[job.uid] = job.status
                 else:
                     raise ValueError("Received unexpected state %s for job with UID %s." % (job.status.name, job.uid))
         self._previous_statuses = new_statuses
+
+    def _timeout(self, entry: DatabaseJobEntry) -> None:
+        work_machine = entry.assigned_machine
+        print("Lost connection to Work machine with id %s: " % work_machine.uid)
```
ammen99: I think you should also set the status of all assigned jobs in the database, which in turns means the dispatcher may need to access the database ...

M1keReck: Also what if the server reconnects to the work machine somehow and the jobs kept running?

src/ja/server/dispatcher/dispatcher.py
```diff
@@ -36,17 +38,35 @@ def set_distribution(self, job_distribution: JobDistribution) -> None:
                 proxy = self._proxy_factory.get_proxy(work_machine)
                 if job.status == JobStatus.RUNNING:
                     if previous_status is None or previous_status == JobStatus.QUEUED:
-                        proxy.dispatch_job(job)
+                        try:
+                            proxy.dispatch_job(job)
+                        except Exception:
```
JohannesGaessler: This catch is too general.
Please replace it with the actual exception time thrown upon timeout.

src/ja/server/dispatcher/dispatcher.py
JohannesGaessler: Should the server periodically try to reconnect to the worker?

Should we introduce an additional JobStatus?
Something like CONNECTION_LOST?

ammen99: I think we can just reuse the CRASHED status, after all the jobs are likely to get aborted when this happens.

ammen99: Would make sense to use the new logging methods, the loggin PR has just been merged.

src/ja/server/dispatcher/dispatcher.py
```diff
@@ -1,9 +1,14 @@
 from ja.common.job import JobStatus
 from ja.server.database.database import ServerDatabase
+from ja.server.database.types.job_entry import DatabaseJobEntry
+from ja.server.database.types.work_machine import WorkMachineState
 from ja.server.dispatcher.proxy_factory import WorkerProxyFactoryBase
 from typing import Dict
+from paramiko.ssh_exception import SSHException
```
M1keReck: This import doesn't work for some reason

JohannesGaessler: I think you mean that mypy complains about the import.
You can fix this by adding `  #type: ignore` after the import.

src/ja/server/dispatcher/dispatcher.py
```diff
@@ -36,17 +41,35 @@ def set_distribution(self, job_distribution: JobDistribution) -> None:
...
+                    except SSHException:
+                        self._timeout(job_entry)
                     new_statuses[job.uid] = job.status
                 elif job.status == JobStatus.QUEUED:
                     new_statuses[job.uid] = job.status
                 else:
                     raise ValueError("Received unexpected state %s for job with UID %s." % (job.status.name, job.uid))
         self._previous_statuses = new_statuses
+
+    def _timeout(self, entry: DatabaseJobEntry) -> None:
```
ammen99: Hmm, does this update the database at all??

Also, shouldn't we update all jobs on the work machine, not just the current one?

M1keReck:
@ammen99 It does now, but the the return type of set_distribution had to be changed.
The Dispatcher sends also a dummy command to all work machines in order to detect offline machines.

src/ja/server/database/sql/database.py
```diff
@@ -270,6 +270,11 @@ def update_work_machine(self, machine: WorkMachine) -> None:
             logger.debug(
                 "old machine: \n %s \n new machine: \n %s" % (str(machine), str(work_machine)))
             work_machine.state = machine.state
+            if work_machine.state == WorkMachineState.OFFLINE:
```
ammen99: This code should be in the scheduler, not in the database.

src/ja/server/scheduler/scheduler.py
```diff
@@ -76,4 +76,7 @@ def reschedule(self, database: ServerDatabase) -> None:
         for machine in available_machines:
             database.update_work_machine(machine)
 
-        self._dispatcher.set_distribution(new_schedule)
+        lost_wms: List[WorkMachine] = self._dispatcher.set_distribution(new_schedule)
```
ammen99: Since this code isn't easily tested, I think you should add at least a unit test. Should be simple - you need:

1. A work machine with 1 job running
2. Mock the dispatcher.set_distribution function to return exactly this work machine
3. Verify that the machine is offline and that the job is crashed

M1keReck:
Second commit fixes #174 by skipping all jobs that are set to be run in the initial iteration through the new job distribution and dispatching/resuming them in the second iteration.

src/ja/server/dispatcher/dispatcher.py
```diff
@@ -22,38 +25,81 @@ def __init__(self, proxy_factory: WorkerProxyFactoryBase):
...
         new_statuses: Dict[str, JobStatus] = dict()
+        self._lost_work_machines: List[WorkMachine] = []
         for job_entry in job_distribution:
             job = job_entry.job
 
             if job.status == JobStatus.QUEUED:
                 logger.debug("Job %s queued." % job.uid)
                 continue
+            if job.status == JobStatus.RUNNING:
+                logger.debug("Job %s is set to run." % job.uid)
```
ammen99: I don't think we need debug's here, since we willl print the same message twice. Better leave only the one below.

Oh, and I have other debug messages already added in the master branch, if you rebase, you can take them (I print the job, status and assigned machine which I think is better)

src/ja/server/dispatcher/dispatcher.py
```diff
@@ -22,38 +25,81 @@ def __init__(self, proxy_factory: WorkerProxyFactoryBase):
...
+                    except SSHException:
+                        self._timeout(job_entry)
                     new_statuses[job.uid] = job.status
                 else:
                     raise ValueError("Received unexpected state %s for job with UID %s." % (job.status.name, job.uid))
+
+        for job_entry in job_distribution:
+            job = job_entry.job
+            if job.status != JobStatus.RUNNING:
+                logger.debug("Job %s is not set to run." % job.uid)
```
ammen99: This should be printed only on queueing.

ammen99:
By the way, sorting won't be a problem since I already sort the distribution once in the scheduler, which means that complexity will remain the same. However, I think sorting makes the logic simpler.

src/ja/server/dispatcher/dispatcher.py
```diff
@@ -22,38 +25,79 @@ def __init__(self, proxy_factory: WorkerProxyFactoryBase):
...
+            if job.status == JobStatus.RUNNING:
+                continue
 
             work_machine = job_entry.assigned_machine
             logger.debug("Job %s with status %s on %s." % (job.uid, job.status.name, work_machine.uid))
+            proxy = self._proxy_factory.get_proxy(work_machine)
+            try:
+                proxy.check_connection()
+            except SSHException:
+                self._timeout(job_entry)
```
JohannesGaessler: Why are you checking the connection to the work machine separately for each job entry?
In the code below there will be a timeout check anyways, right?

ammen99:
I think we should separate the changes in this PR in two parts, and prioritize the one with sorting job entries before dispatching (Otherwise, we may get thrashing on one worker if one of the commands is CANCEL, or maybe the command will fail because it tries to acquire a license which isn't freed yet)

src/ja/server/dispatcher/dispatcher.py
M1keReck: If the job status didn't change, there won't be a check. So if the server loses connection to a work machine that's running a job, the server keeps storing it as a running job indefinitely.

ammen99:
@M1keReck I have included a few fixes in the timeout-2 branch, I think you should include them. For example:

1. Changed it so that you send the dummy command only to jobs which you don't check otherwise.
2. I made it so that the dispatcher doesn't return multiple machines
3. I made it so that you only set the running jobs to crashed, not old jobs.

ammen99:
I think you need to rebase.

M1keReck:
[Build](https://travis-ci.org/github/DistributedTaskScheduling/JobAdder/builds/668779125?utm_medium=notification&utm_source=email)

Issue #113, opened 2020-03-03
Infrastructure for integration phase


ammen99:
Class which starts all 3 components.

PR #122, opened 2020-03-08
Integration infrastructure

JohannesGaessler:
Fixes https://github.com/DistributedTaskScheduling/JobAdder/issues/113.

I have done relatively little work on this PR because I was waiting for more progress to be done in https://github.com/DistributedTaskScheduling/JobAdder/pull/108 and https://github.com/DistributedTaskScheduling/JobAdder/pull/112.
To make the worker main usable I applied some quick, hacky fixes.
On the worker main PR the JobWorker argument in the execute methods should be replaced with DockerInterface.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,21 @@ language: python
 python: 3.7
 install: "pip3 install flake8 mypy==0.761 PyYAML psycopg2 SQLAlchemy testing.postgresql sqlalchemy-stubs paramiko freezegun docker"
 script:
+  - sudo groupadd jobadder
+  - sudo usermod -a -G jobadder travis
+  - ssh-keygen -f ~/.ssh/id_rsa -P "" -t rsa -m PEM
+  - cp ~/.ssh/id_rsa.pub ~/.ssh/authorized_keys
+  - echo PYTHONPATH=$PWD/src >> ~/.ssh/environment
+  - echo "PermitUserEnvironment yes" | sudo tee -a /etc/ssh/sshd_config
+  - sudo systemctl reload sshd
+  - sudo -u postgres createuser jobadder
+  - sudo -u postgres psql -c "ALTER USER jobadder PASSWORD 'jobadder';"
+  - sudo -u postgres createdb jobadder-test
   - flake8 src/ --ignore F821 --max-line-length=120
   - ./.travis/check_init_py.sh src/
   - MYPYPATH=$(pwd)/src mypy -p ja --no-strict-optional --strict --config-file $(pwd)/mypy.ini
   - MYPYPATH=$(pwd)/src mypy -p test --no-strict-optional --strict
-  - cd src && ./setup.py test
+  - cd src && python3 -m unittest discover --verbose --start-directory test --pattern "*.py"
 services:
   - docker
+  - postgresql
--- a/src/ja/common/message/server.py
+++ b/src/ja/common/message/server.py
@@ -2,10 +2,11 @@
 This module provides the interfaces for Messages sent to the central server.
 """
 from abc import ABC, abstractmethod
-from typing import cast, Dict
+from typing import cast, Dict, TYPE_CHECKING
 
 from ja.common.message.base import Command, Response
-from ja.server.database.database import ServerDatabase
+if TYPE_CHECKING:
+    from ja.server.database.database import ServerDatabase
 
 
 class ServerCommand(Command, ABC):
@@ -16,7 +17,7 @@ class ServerCommand(Command, ABC):
     """
 
     @abstractmethod
-    def execute(self, database: ServerDatabase) -> Response:
+    def execute(self, database: "ServerDatabase") -> Response:
         """!
         Executes a ServerCommand object on the server and generates a
         ServerResponse object to be sent back in return.
--- a/src/ja/common/proxy/ssh.py
+++ b/src/ja/common/proxy/ssh.py
@@ -1,6 +1,7 @@
 from typing import Dict
 from abc import ABC, abstractmethod
 from paramiko import SSHClient, AutoAddPolicy  # type: ignore
+import yaml
 
 from ja.common.message.server import ServerCommand, ServerResponse
 from ja.common.message.worker import WorkerCommand, WorkerResponse
@@ -42,7 +43,7 @@ class ISSHConnection(ABC):
 
 
 class SSHConnection(ISSHConnection):
-    def __init__(self, ssh_config: "SSHConfig", remote_module: str):
+    def __init__(self, ssh_config: "SSHConfig", remote_module: str, command_string: str = "python3 -m %s"):
         """!
         Creates a new SSHConnection object. Arguments for establishing the
         actual ssh connection are packaged in @ssh_config. If no credentials
@@ -50,6 +51,7 @@ class SSHConnection(ISSHConnection):
         constructor prompts the user for input.
         @param ssh_config: Config for paramiko.
         @param remote_module: The Python module to execute on the host.
+        @param command_string: the template for executing commands on the remote component.
         """
         self._username = ssh_config.username
         self._client = SSHClient()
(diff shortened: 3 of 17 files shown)


nikolatzotchev:
I have a few questions

1.  I was getting some errors with the database, I could log in without having a password to the user, I just set up 1234 and it works, perhaps it has something to do with the fact that I have edited the pg_hba.conf, can you please show me yours so that I can copy it.

2. Now the stuff with the database works, but I'm getting `paramiko.ssh_exception.NoValidConnectionsError: [Errno None] Unable to connect to port 22 on 127.0.0.1 or ::1
` Do you have any clue what is wrong.

.travis.yml
```diff
@@ -2,10 +2,20 @@ language: python
 python: 3.7
 install: "pip3 install flake8 mypy==0.761 PyYAML psycopg2 SQLAlchemy testing.postgresql sqlalchemy-stubs paramiko freezegun docker"
 script:
+  - sudo groupadd jobadder
+  - sudo usermod -a -G jobadder travis
+  - ssh-keygen -f ~/.ssh/id_rsa -P ""
```
ammen99: Please specify more clearly that we want an RSA key in a specific format, otherwise on some very recent distributions like F31 the generated key is not recognized as valid by paramiko. We need:

`ssh-keygen -f ~/.ssh/id_rsa -P "" -t rsa -m PEM`

ammen99:
I have tried to start this and got the same problems as @nikolatzotchev . I found what the problem is:

We are using `test.integration....`. However, it seems that the name `test` can conflict with https://docs.python.org/3/library/test.html and if both are present, then python thinks we are trying to access something in this library.

As a workaround, I suggest that we move the integration tests to a third folder `src/ja_integration` so that we are sure we don't have any such conflicts.

src/ja/server/main.py
```diff
@@ -32,21 +32,24 @@ def _read_config() -> ServerConfig:
...
         """
 
         config = self._read_config()
         self._database = SQLDatabase(host=config.database_config.host,
                                      user=config.database_config.username,
-                                     password=config.database_config.password)
-        proxy_factory = WorkerProxyFactory(self._database)
+                                     password=config.database_config.password,
+                                     database_name=database_name)
+        proxy_factory = self._get_proxy_factory()
```
ammen99: I thought the goal is to test with all real components, including the real proxy factory. Why do we need to change that?

src/ja/worker/main.py
```diff
@@ -15,20 +15,28 @@ class JobWorker:
     """
     The main class on the Worker Client
     """
-    def __init__(self, config_path: str = "/etc/jobadder/worker.conf") -> None:
+    def __init__(
+            self, config_path: str = "/etc/jobadder/worker.conf",
+            socket_path: str = "/run/jobadder-worker.socket/",
```
ammen99: same thing as the other PR, why `/` at the end?

src/ja/worker/proxy/command_handler.py
```diff
@@ -21,8 +21,10 @@ class WorkerCommandHandler(CommandHandler):
     The command handler creates a socket.
     It starts listening on it for incoming commands.
     """
-    def __init__(self, admin_group: str, docker_interface: DockerInterface):
-        super().__init__(socket_path="/run/jobadder-worker.socket/", admin_group=admin_group)
+    def __init__(self,
+                 admin_group: str, docker_interface: DockerInterface,
+                 socket_path: str = "/run/jobadder-worker.socket/"):
```
ammen99: `/` at the end.

src/test/integration/base.py
```diff
@@ -0,0 +1,141 @@
...
+        sleep(1)
+
+    def tearDown(self) -> None:
+        database = self._server._database
+        for table in reversed(database._metadata.sorted_tables):
+            database.engine.execute(table.delete())
+
+    @property
+    def ssh_config(self) -> SSHConfig:
+        return SSHConfig(hostname="localhost", username=getuser())
```
ammen99: using 127.0.0.1 is more versatile, I had 127.0.0.1 working on my laptop but localhost not. It is all a matter of configuration in the end, but 127.0.0.1 should work in more cases.

src/ja/server/main.py
ammen99: Hmm, is it because you want to use Proxies which use the custom Remotes ?

src/ja/common/message/server.py
```diff
@@ -2,10 +2,11 @@
 This module provides the interfaces for Messages sent to the central server.
 """
 from abc import ABC, abstractmethod
-from typing import cast, Dict
+from typing import cast, Dict, TYPE_CHECKING
 
 from ja.common.message.base import Command, Response
-from ja.server.database.database import ServerDatabase
+if TYPE_CHECKING:
```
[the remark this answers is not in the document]
JohannesGaessler: I removed the WorkerConnectionDetails class and replaced it with SSHConnection.
This resulted in a circular import and I judged this position to be the best to fix it.

src/ja/common/proxy/ssh.py
```diff
@@ -161,6 +166,8 @@ def to_dict(self) -> Dict[str, object]:
 
     @classmethod
     def from_dict(cls, property_dict: Dict[str, object]) -> "SSHConfig":
+        if property_dict is None:
```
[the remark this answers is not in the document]
JohannesGaessler: There were multiple classes where a property of type SSHConnection could be None.
I figured that instead of checking externally if the dict is None this would be simpler.

src/ja/server/database/sql/database.py
```diff
@@ -20,143 +20,142 @@ class SQLDatabase(ServerDatabase):
...
 
         @param host The host of the database to connect to.
         @param user The username to use for the connection.
         @param password The password to use for the connection.
+        @param database_name The name of the database to use.
         """
         self.scheduler_callback: Callable[["ServerDatabase"], None] = lambda *args: None
         self.status_callback: Callable[["Job"], None] = lambda *args: None
         self.in_scheduler_callback: bool = False
-        if SQLDatabase._metadata:
```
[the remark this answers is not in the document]
JohannesGaessler: Because otherwise you could only create one instance of SQLDatabase per process.
Below the initialization of SQLDatabase._metadata there are object properties being defined.

src/ja/server/main.py
JohannesGaessler: The default locations for configs and sockets are only accessible by root.
I added subclasses of the main classes and some other classes that define different paths for configs and sockets.
This is to allow non-root users to run the integration tests and to prevent unintentional changes to the system (just imagine a bug like "rm -r /etc /jobadder" instead of "rm -r /etc/jobadder").

src/ja/worker/main.py
JohannesGaessler: Because I copy-pasted the typo.

src/ja/common/proxy/ssh.py
```diff
@@ -59,21 +61,24 @@ def __init__(self, ssh_config: "SSHConfig", remote_module: str):
...
         )
         self._remote_module = remote_module
+        self._command_string = command_string
 
     def send_server_command(self, command: ServerCommand) -> ServerResponse:
-        stdin, stdout, stderr = self._client.exec_command("python3 -m %s" % self._remote_module)
-        stdin.write(command)
-        response = ServerResponse.from_string(stdout.read())
+        command_dict = dict(command=command.to_dict(), type_name=command.__class__.__name__)
+        stdin, stdout, stderr = self._client.exec_command(self._command_string % self._remote_module)
```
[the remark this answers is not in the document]
nikolatzotchev: I will create a separate PRs, for the logging in user, workers and common.

JohannesGaessler:
I pushed a new version where I moved the remotes to ja_integration.remote.
Please confirm whether this works on your machine @ammen99 @nikolatzotchev .

nikolatzotchev:
It works on my pc.

ammen99:
It doesn't on mine :( It still complains about the `import test...`)

JohannesGaessler:
@ammen99 Just to make sure: Did you forget to uninstall JobAdder after you tested the pip installation?

ammen99:
Yes, I just went through all of the places where pip could install job adder and there is nothing installed.

ammen99:
Also, please add a password to the database. My postgresql installation does not want to accept connections without password.

JohannesGaessler:
>My postgresql installation does not want to accept connections without password.

Does setting the password in line 105 of test/integration/base.py not work?

ammen99:
> Does setting the password in line 105 of test/integration/base.py not work?

Yes, but I have to do it manually. I think it will not hurt to add it to the PR for ease of testing on our computers (Nikola had the same issue). Plus I assume we're going to write more tests using this infrastructure so we'll need to do that in the future, on our PCs.

JohannesGaessler:
>It doesn't on mine :( It still complains about the import test...)

Please try again with the newest version.

ammen99:
> > 
> 
> Please try again with the newest version.

Newest version of what?

JohannesGaessler:
Forgot to push it, sry.

ammen99:
@JohannesGaessler Yes, the last commit works when I add the password manually.

src/ja/server/main.py
ammen99: Yeah, makes sense.

src/ja/common/proxy/ssh.py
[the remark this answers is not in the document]
ammen99: ok, then you'll have to add this.

src/ja/server/database/sql/database.py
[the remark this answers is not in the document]
ammen99: Oh sorry, I somehow thought it was working before, that's why I didn't get why you need to change it.

One more question, out of curiosity, when do we need multiple databases per process?

JohannesGaessler:
>One more question, out of curiosity, when do we need multiple databases per process?

When doing the integration tests :^) .
The setUp method creates a new instance of JobCenter which in turn creates a new instance of SQLDatabase.

ammen99:
> When doing the integration tests :^) .
> The setUp method creates a new instance of JobCenter which in turn creates a new instance of SQLDatabase.

Isn't that like one database here? And then you can get JobCenter._database instead of creating a new one.

Doesn't matter of course.

Can you rebase this PR so we can merge it?

JohannesGaessler:
I excluded ja_integration in the install script.
Is there something else that I would need to consider for your changes @ammen99 ?

ammen99:
Good catch. I do not think there is anything else special.

PR #129, opened 2020-03-11
fixing dict

### LINKED ISSUE #118: Augment proxies to set type_name
I just noticed we never set the `type_name` which the command handlers use to differentiate between commands:

https://github.com/DistributedTaskScheduling/JobAdder/blob/master/src/ja/common/proxy/command_handler.py#L60

As a result, the server and the worker will never be able to parse commands (except if I am mistaken and this is set somewhere, but I could not find any reference to `type_name` aside from the command handlers). I think we can either

1. Make the proxies pass it as an additional argument to the SSH connection

2. Make the SSH connection infer the type by looking at the Python class name (`object.__class__.__name__`)

What do you think is the better approach? My personal preference is 2. since it is easier to implement and won't require changing the interfaces, plus removes the need to manually specify the type everywhere.

nikolatzotchev:
fixes #118 probably, I noticed that `username` is also never set.

### DIFF
--- a/src/ja/common/proxy/ssh.py
+++ b/src/ja/common/proxy/ssh.py
@@ -2,7 +2,8 @@ from typing import Dict
 from abc import ABC, abstractmethod
 from paramiko import SSHClient, AutoAddPolicy  # type: ignore
 
-from ja.common.message.server import ServerCommand, ServerResponse
+from ja.common.message.base import Response, Command
+from ja.common.message.server import ServerCommand
 from ja.common.message.worker import WorkerCommand, WorkerResponse
 from ja.common.config import Config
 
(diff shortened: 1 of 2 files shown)


JohannesGaessler:
During my implementation of https://github.com/DistributedTaskScheduling/JobAdder/pull/122 I am already fixing this issue.
This is in addition to other fixes for SSHConnection like the order of read/write operations resulting in deadlocks.
I think the easiest approach going forward would be to close this PR and for me to integrate the other changes made here into my PR.

nikolatzotchev:
Sorry for some reason I didn't get and github notification for this commend will close it down.

PR #130, opened 2020-03-15
Setting version for mypy.

nikolatzotchev:
This should fix the problems with mypy.


JohannesGaessler:
Which problems specifically?

nikolatzotchev:
The newer version marks some stuff in the database as errors (0.770). I should either add typing: ignore to all the lines, or just downgrade mypy, You can check [this](https://github.com/DistributedTaskScheduling/JobAdder/pull/116) failed build log.

PR #134, opened 2020-03-22
Removed WorkerResponse and ServerResponse

### LINKED ISSUE #95: Response classes should be unified
As discussed in the meting, I think we should have only a single Response class instead of Response, ServerResponse, and WorkerResponse.
Since we currently have a lot of PRs open this change should wait until we have a lower amount of PRs open.

nikolatzotchev:
fixes #95


src/ja/common/proxy/ssh.py
```diff
@@ -16,21 +17,21 @@ class ISSHConnection(ABC):
     """
 
     @abstractmethod
-    def send_server_command(self, command: ServerCommand) -> ServerResponse:
+    def send_server_command(self, command: ServerCommand) -> Response:
```
JohannesGaessler: Now that there is only one Response class, please unify send_server_command and send_worker_command.
I only created two separate methods due to the limitations imposed imposed by mypy.

src/ja/common/proxy/ssh.py
```diff
@@ -17,21 +16,12 @@ class ISSHConnection(ABC):
     """
 
     @abstractmethod
-    def send_server_command(self, command: ServerCommand) -> ServerResponse:
+    def send_command(self, command: Command) -> Response:
         """!
         Sends a ServerCommand to the Remote on the host. Automatically attaches the username specified in ssh_config to
```
ammen99: Sends a Command*

src/ja/common/proxy/ssh.py
```diff
@@ -17,21 +16,12 @@ class ISSHConnection(ABC):
     """
 
     @abstractmethod
-    def send_server_command(self, command: ServerCommand) -> ServerResponse:
+    def send_command(self, command: Command) -> Response:
         """!
-        Sends a ServerCommand to the Remote on the host. Automatically attaches the username specified in ssh_config to
+        Sends a Command to the Remote on the host. Automatically attaches the username specified in ssh_config to
         the ServerCommand.
```
ammen99: still a ServerCommand here :P

nikolatzotchev:
will do

ammen99:
needs a rebase

PR #135, opened 2020-03-22
tests for user commands

### LINKED ISSUE #133: Improve tests for user commands
Right now, we have no tests for the `execute()` methods of user commands, and they have a lower coverage than most of the code.

M1keReck:
fixes #133

### DIFF
--- a/src/test/user/cli.py
+++ b/src/test/user/cli.py
@@ -3,6 +3,7 @@ from typing import List, cast
 
 from ja.common.job import Job, JobPriority, JobSchedulingConstraints, JobStatus
 from ja.common.docker_context import DockerConstraints, DockerContext, MountPoint
+from ja.server.database.sql.mock_database import MockDatabase
 from ja.user.message.add import AddCommand
 from ja.user.message.query import QueryCommand
 from ja.user.message.cancel import CancelCommand
@@ -19,7 +20,7 @@ class CLITest(TestCase):
                                  "add -s test/user/program.py -l lab -e a@e.de -p low -t 1 -m 2 --owner 9".split(),
                                  "add -s test/user/program.py -e a@e.de -p low -t 6 -m 4 --owner 9".split(),
                                  "add -s test/user/program.py -e a@e.de -p high -t 1 -m 4 --owner 9".split(),
-                                 "add -c test/user/add.yaml".split()]
+                                 "add -c test/user/add.yaml --owner 0".split()]
 
     def setUp(self) -> None:
         self._jobs: List[Job] = [Job(1, "anon@biz.org", JobSchedulingConstraints(JobPriority.HIGH, False, []),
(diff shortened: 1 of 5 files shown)


src/ja/server/database/database.py
```diff
@@ -20,15 +20,6 @@ def find_job_by_id(self, job_id: str) -> DatabaseJobEntry:
         @return The stored data about the job, or None if no such job was found.
         """
 
-    @abstractmethod
```
ammen99: Why did you remove `find_job_by_label`, and then you reimplement it using `query_jobs` and manually filter through them?

src/test/user/cli.py
```diff
@@ -46,70 +48,106 @@ def setUp(self) -> None:
...
         command: QueryCommand = cast(QueryCommand, self._cli.get_server_command("query --sr lic,lic1".split()))
         response: str = self._proxy.query(command).result_string
+        response1: str = command.execute(self._db).result_string
         self.assertEqual(response, "\n".join([str(self._jobs[4])]))
+        self.assertEqual(response1, "\n".join([str(self._jobs[4])]))
 
     def test_cancel_label(self) -> None:
         command: CancelCommand = cast(CancelCommand, self._cli.get_server_command("cancel --label lab".split()))
         self._proxy.cancel_job(command)
+        command.execute(self._db).result_string
```
nikolatzotchev: you forgot to remove .result_string (or forgot to assign it)

src/test/user/cli.py
```diff
@@ -46,70 +48,106 @@ def setUp(self) -> None:
...
+        self.assertEqual(response1, "\n".join([str(self._jobs[1])]))
 
     def test_query_memory_empty(self) -> None:
         command: QueryCommand = cast(QueryCommand, self._cli.get_server_command("query --memory 10 40".split()))
         response: str = self._proxy.query(command).result_string
+        response1: str = command.execute(self._db).result_string
         self.assertEqual(response, "No jobs satisfy these constraints.")
+        self.assertEqual(response1, "No jobs satisfy these constraints.")
 
     def test_query_sresources(self) -> None:
```
nikolatzotchev: sresources -> resources

ammen99:
I think we have missed this when doing the previous reviews. The tests for commands should be separate, not part of the CLI tests.

src/ja/server/database/database.py
M1keReck: Because I couldn't make it work with `session.query(Job).filter(Job.label == label)`. The method didn't work.

src/test/user/cli.py
M1keReck: That stands for special resources.

src/ja/server/database/database.py
JohannesGaessler: I opened a PR that fixes the method: https://github.com/DistributedTaskScheduling/JobAdder/pull/135

ammen99: That PR is now merged, so it makes the most sense to use the database function.

src/ja/server/database/sql/database.py
```diff
@@ -196,6 +196,7 @@ def update_job(self, job: Job) -> None:
                                                                     0, 0),
                                          machine=None)
             session.add(job_entry)
+            # job.uid =
```
M1keReck: How could I get the id of the newly added job entry before adding it? Since the id has to be assigned to the job before adding it to the DB

nikolatzotchev: the id is assigned automatically, uid is not, you can't get the id before adding.

ammen99:
@M1keReck It is generally better to rebase on top of master instead of merging branches.

ammen99:
I manually re-triggered the TravisCI build because I suspected we'll have failing tests.. However I think that maybe you just need to rebase on top of master.

PR #137, opened 2020-03-23
messages: add effective_user for user messages

### LINKED ISSUE #126: Add effective_user parameter to UserServerCommands
The user -> server commands need to verify that the effective user has permission to carry out the request. The server command handler should therefore give this effective user to the command (we need to add a new parameter to the execute() method)

ammen99:
Fixes #126

I decided it makes the most sense to add a base class for all user messages where I can set the effective user for each command. This seemed the easiest way to add it, without having to change many tests.

### DIFF
--- a/src/ja/user/message/add.py
+++ b/src/ja/user/message/add.py
@@ -33,6 +33,9 @@ class AddCommand(UserServerCommand):
 
     def execute(self, database: ServerDatabase) -> Response:
         job: Job = self.config.job
+        if job.owner_id == -1:
+            job.owner_id = self.effective_user
+
         if self.effective_user != 0 and self.effective_user != job.owner_id:
             return Response(result_string="Cannot submit jobs for other users.", is_success=False)
 
--- a/src/ja/user/message/add.py
+++ b/src/ja/user/message/add.py
@@ -1,5 +1,5 @@
 from typing import Dict
-from ja.common.message.server import ServerCommand
+from ja.user.message.base import UserServerCommand
 from ja.common.message.base import Response
 from ja.server.database.database import ServerDatabase
 from ja.user.config.add import AddCommandConfig
@@ -7,7 +7,7 @@ from ja.common.job import Job, JobStatus
 from ja.server.database.types.job_entry import DatabaseJobEntry
 
 
-class AddCommand(ServerCommand):
+class AddCommand(UserServerCommand):
     """
     Command for adding a job.
     """
@@ -33,6 +33,9 @@ class AddCommand(ServerCommand):
 
     def execute(self, database: ServerDatabase) -> Response:
         job: Job = self.config.job
+        if self.effective_user != 0 and self.effective_user != job.owner_id:
+            return Response(result_string="Cannot submit jobs for other users.", is_success=False)
+
         db_job_id: DatabaseJobEntry = database.find_job_by_id(job.uid)
         if db_job_id is not None:
             return Response(result_string="Job with id %s already exists" % job.uid,
--- a/src/ja/user/message/base.py
+++ b/src/ja/user/message/base.py
@@ -1,0 +1,14 @@
+from ja.common.message.server import ServerCommand
+
+
+class UserServerCommand(ServerCommand):
+    def __init__(self) -> None:
+        self._effective_user: int = -1
+
+    @property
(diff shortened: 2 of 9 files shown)


src/ja/user/message/base.py
```diff
@@ -0,0 +1,14 @@
+from ja.common.message.server import ServerCommand
+
+
+class UserServerCommand(ServerCommand):
+    def __init__(self) -> None:
+        self._effective_user: int = -1
+
+    @property
+    def effective_user(self) -> int:
```
JohannesGaessler: Why are you implementing this property twice?

src/ja/user/message/cancel.py
```diff
@@ -68,6 +69,15 @@ def from_dict(cls, property_dict: Dict[str, object]) -> "CancelCommand":
         cls._assert_all_properties_used(property_dict)
         return CancelCommand(label=_label, uid=_uid, config=_config)
 
+    def _cancel_job(self, database: ServerDatabase, job: Job, ignore_wrong_permissions: bool) -> bool:
+        has_permissions = (job.uid == self.effective_user or self.effective_user == 0)
```
JohannesGaessler: I think this should be job.owner_id instead of job.uid.

src/ja/user/message/add.py
```diff
@@ -33,6 +33,9 @@ def from_dict(cls, property_dict: Dict[str, object]) -> "AddCommand":
 
     def execute(self, database: ServerDatabase) -> Response:
         job: Job = self.config.job
+        if self.effective_user != 0 and self.effective_user != job.owner_id:
+            return Response(result_string="Cannot submit jobs for other users.", is_success=False)
```
JohannesGaessler: Shouldn't the command just set job.owner_id to self.effective_user?
I don't think Job.owner_id is ever changed from its default of -1 (outside of tests).

nikolatzotchev: wouldn't the command handler set the effective_user at line 80
`        user_command.effective_user = 0 if self._user_is_admin(user) else pwd.getpwnam(user).pw_uid`

ammen99: @JohannesGaessler  I do not think this is the case. My main consideration was that, at some point in the future, there might be a need for administrators to restart jobs or something like that. In which case, administrators would need to be able to submit jobs for other people.

JohannesGaessler: I think you are not understanding the problem:
The Job class has an owner_id property.
As of this PR commands also have an effective_user property.
For regular users permission checks are performed by comparing these two properties.
However, as far as I can tell we haven't yet implemented setting Job.owner_id.
Job.owner_id is initialized with -1 in the CLI handler and never changed from this default after that.
In any test that we did we always set Job.owner_id manually.
I think that if we do not change this regular users will not be able to add jobs at all.

If we want to allow admins to add jobs for other users I would suggest something like this:
If `Job.owner_id == -1` then `Job.owner_id = Command.effective_user` else `check_permissions()`.

ammen99: The proper fix would be to set the owner id in the user CLI, not here.

JohannesGaessler: No it wouldn't be.
The user client and the server are by design running on different systems.
Linux username and user IDs are generally **not** the same across systems.

ammen99: Didn't we say we have LDAP and the same users everywhere?

JohannesGaessler: The way I remember it is that we would have the same users on the server and the workers but **not** on the user clients.

ammen99: Okay, I had another impression since we were talking about mounting the user's home directory or something similar by default. This is not possible if the users are different, but whatever, should be fixed now.

JohannesGaessler: Well, if we were to mount the home directory of a user it would be the home directory of the user that is used for the SSH login, in other words effective_user, right?

JohannesGaessler: >This is not possible if the users are different

Why?

ammen99: Let's not waste time on this.

JohannesGaessler: Okay.

PR #138, opened 2020-03-23
Impl proxy factory

### LINKED ISSUE #121: Set up remote executables
We need to figure out how to run the correct remote (server vs worker) and of course install them as scripts with SUID.

Should probably be based on #119 since it tries to set up an installation framework.

### LINKED ISSUE #125: Implement WorkerProxyFactory._create_proxy

ammen99:
Fixes #125 
Fixes #121

I had to add an external C program so that we can SUID the Remotes, otherwise it will not work since SUID for scripts is disabled in the kernel(for a good reason!), and using a non-safe SUID wrapper is a security risk.

### DIFF
--- a/src/ja/server/dispatcher/proxy_factory.py
+++ b/src/ja/server/dispatcher/proxy_factory.py
@@ -3,7 +3,7 @@ from typing import Dict
 
 from ja.server.database.database import ServerDatabase
 from ja.server.database.types.work_machine import WorkMachine
-from ja.server.proxy.proxy import IWorkerProxy
+from ja.server.proxy.proxy import IWorkerProxy, WorkerProxy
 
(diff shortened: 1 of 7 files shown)


src/ja/server/dispatcher/proxy_factory.py
```diff
@@ -45,4 +45,5 @@ class WorkerProxyFactory(WorkerProxyFactoryBase):
     """
 
     def _create_proxy(self, work_machine: WorkMachine) -> IWorkerProxy:
+        return WorkerProxy(work_machine.uid, work_machine.ssh_config)
         pass
```
JohannesGaessler: I think you forgot to remove the pass token.

PR #140, opened 2020-03-24
Fix for querying by label

JohannesGaessler:
Querying by label is broken on master.
This PR fixes the database method and adds a unittest for it.


ammen99:
Since Travis wasn't working again we merged this PR even though the build was failing .. I force-pushed a fix directly to master since the fix was trivial (just ` -> None`)

Issue #142, opened 2020-03-25
get_work_machines() in the SQLDatabase returns all machines


ammen99:
The docstring returns only online machines, and the rest of the code assumes the docstring is correct.

PR #143, opened 2020-03-25
Worker safe shutdown

### LINKED ISSUE #127: Graceful worker shutdown

ammen99:
Fixes #127 

~~I tried to test server+worker locally, but I encountered #142, which makes it impossible for a worker to register itself again after being registered in the database more than once ...~~ Now that bug is fixed, but another one has come around ..

### DIFF
--- a/data/bin/ja-server
+++ b/data/bin/ja-server
@@ -1,32 +1,17 @@
 #!/usr/bin/env python3
 
-from argparse import ArgumentParser
-from io import StringIO
-from ja.common.message.base import Response
-from ja.common.proxy.remote import Remote
+from ja.common.cli_util import SimpleDaemonCLIInterface
 from ja.server.main import JobCenter
-from typing import cast
-import yaml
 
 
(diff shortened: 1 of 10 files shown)


ammen99:
I think shut down == retire, so we are waiting for the jobs to get finished, right?

JohannesGaessler:
It would maybe be nice to have a --force option that just cancels all jobs but I don't think it's a priority.

ammen99:
> It would maybe be nice to have a --force option that just cancels all jobs but I don't think it's a priority.

I also don't think it is very important since we still have crashes to fix like #145, plus I don't think we have tested the scheduler/dispatcher at all so far ...

ammen99:
@JohannesGaessler Should I take your comment as an approving review?

ammen99:
> I didn't approve because there was talk in the OP about a bug but I assume it's fixed then?

Oh, I should have clarified. The bug is when testing the worker, not directly related or caused by this PR.

PR #144, opened 2020-03-25
Fixed bug in database.

nikolatzotchev:
get_work_machines was returning all machines not only the ones that were ONLINE.


src/ja/server/database/sql/database.py
```diff
@@ -265,7 +266,8 @@ def update_work_machine(self, machine: WorkMachine) -> None:
 
     def get_work_machines(self) -> Optional[List[WorkMachine]]:
         session = self.scoped()
-        work_machines: Optional[List[WorkMachine]] = session.query(WorkMachine).options(joinedload("*")).all()
+        work_machines: Optional[List[WorkMachine]] = session.query(WorkMachine).filter(
+            WorkMachine.state != WorkMachineState.OFFLINE).options(joinedload("*")).all()
```
ammen99: Could also be nice to specify retired machines are also included in the docstring in database.py

PR #147, opened 2020-03-26
user-cli: set owner_id to current user if not set on the cmdline

ammen99:
This should be done here, not in the execute() method.


ammen99:
Seems like this is not the correct fix after all.

PR #150, opened 2020-03-26
Save ssh_config to the database.

### LINKED ISSUE #141: How does the server know the appropriate SSH connection for the work machine?
I couldn't find the code, and I assume it is missing. I am also not sure what our decision to implement this was.

### LINKED ISSUE #149: Missing SSHConfig in the database
We need a proper table for that.

nikolatzotchev:
fixes #149 
fixes #141
Somehow I forgot to add this.
Fixed every test so that the work machines have ssh_config

### DIFF
--- a/src/ja/worker/main.py
+++ b/src/ja/worker/main.py
@@ -20,6 +20,7 @@ class JobWorker:
     """
     The main class on the Worker Client
     """
+
     def __init__(
             self, config_path: str = "/etc/jobadder/worker.conf",
             socket_path: str = "/tmp/jobadder-worker.socket",
--- a/src/test/server/database/test_database.py
+++ b/src/test/server/database/test_database.py
@@ -6,6 +6,7 @@ from ja.server.database.sql.mock_database import MockDatabase
 from ja.common.job import Job, JobSchedulingConstraints, JobPriority, JobStatus
 from ja.common.docker_context import DockerContext, MountPoint, DockerConstraints
 from ja.server.database.types.work_machine import WorkMachine, WorkMachineState, WorkMachineResources
+from ja.common.proxy.ssh import SSHConfig
 import time
 
 from ja.server.database.database import ServerDatabase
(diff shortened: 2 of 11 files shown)


ammen99:
This was forgotten in the PR which actuall added ssh config instead of work machine connection info, don't remember which one was that, but I know we added this after the database.

nikolatzotchev:
Doest anyone have an idea how to fix this error.

nikolatzotchev:
It looks like the tables are not being created in the integration test, no idea how it works in the mock, but doesn't in the real.

nikolatzotchev:
wtf ok now the error is only local ¯\_(ツ)_/¯

nikolatzotchev:
Ok I'm just retarded.

src/ja/worker/main.py
```diff
@@ -43,7 +44,8 @@ def run(self) -> None:
         Registers this worker on the central server and starts the  main loop of the command handler.
         """
         register_response = self._server_proxy.register_self(
-            uid=self._config.uid, work_machine_resources=WorkMachineResources(self._config.resources))
+            uid=self._config.uid, work_machine_resources=WorkMachineResources(self._config.resources),
+            ssh_config=self._config.ssh_config)
```
ammen99: This config is wrong, and does not belong to this PR. I have assigned myself for the task and I'll submit a PR where this is fixed, so please remote it from here.

nikolatzotchev: will do

ammen99: or, just cherry-pick my commit and I'll close my PR. I realize that both are dependent on each other.

src/test/server/database/test_database.py
```diff
@@ -49,11 +50,22 @@ def setUp(self) -> None:
         )
         self.job2.uid = "asv"
         self.work_machine = WorkMachine("machi1", WorkMachineState.ONLINE,
-                                        WorkMachineResources(ResourceAllocation(12, 32, 12)))
+                                        WorkMachineResources(ResourceAllocation(12, 32, 12)),
+                                        SSHConfig(
+                                            hostname="om", username="ux", password="45", key_filename="rsa",
```
ammen99: Can the database work with some of those set to None? I think we do not have all of the fields, all of the time.

src/test/server/database/types/work_machine.py
```diff
@@ -47,9 +48,13 @@ def test_resources_getter1(self) -> None:
 
 class WorkMachineSerializableTest(AbstractSerializableTest):
     def setUp(self) -> None:
-        self._optional_properties = ["resources"]
+        self._optional_properties = ["resources", "ssh_config"]
```
ammen99: My understanding was that ssh_config is no longer optional?

src/test/server/database/test_database.py
nikolatzotchev: the database for sure can, I just copy pasted them this way, also it can always store None, the problem was with using to_dict.

src/test/server/database/types/work_machine.py
nikolatzotchev: Yea but it had a default value of None that is why I put it there, should I remove None as default?

src/test/server/database/test_database.py
ammen99: I think it would still be nice to have it included in some tests. Maybe you could set a few of these here to None?

src/test/server/database/types/work_machine.py
ammen99: If it is mandatory, then I think it should not be part of the `optional_properties` :)

ammen99: Also mandatory == no need for default value afaiu

nikolatzotchev: But that is also true for resources?

ammen99: Then I'm not sure why these are optional either?

nikolatzotchev: yea but maybe for tests, because now if I try to fix something I get 54 failed tests.

JohannesGaessler: When a property is mandatory it has to be specified when creating the object.
When a property is None, that is treated as the property not being specified.
self._optional_properties is a list of properties that are not mandatory.

JohannesGaessler: I think ssh_config should not be allowed to be None.
The from_dict method will need to be adjusted.

nikolatzotchev:
should I try refactoring everything or should I leave it as so. I'm talking about WorkMachine default values.

JohannesGaessler:
I think since we're short on time we should ultimately leave it as-is; the opportunity cost of changing things is too high.

Issue #154, opened 2020-03-27
User client main class


JohannesGaessler:
It seems we never implemented the user client main class (no one was ever assigned).

PR #156, opened 2020-03-27
User client main class

JohannesGaessler:
Fixes https://github.com/DistributedTaskScheduling/JobAdder/issues/154 .

This PR adds a main class for the user client + an executable file called "jobadder" to use on the user client.
Some of the components on the user client were incompatible with each other and had to be adjusted.
I did not implement any unittests for the user main class.


src/ja/user/main.py
```diff
@@ -1,4 +1,35 @@
...
     """
     The main class for the JobAdder user client.
     """
+    def __init__(self, config_path: str = "~/.config/jobadder", remote_module: str = "ja.server.proxy.remote",
+                 command_string: str = "python3 -m %s") -> None:
+        self._cli_handler = UserClientCLIHandler(config_path=config_path)
+        self._remote_module = remote_module
+        self._command_string = command_string
+
+    def run(self, cli_args: List[str] = None) -> None:
```
ammen99: what is the purpose of the `cli_args`?

JohannesGaessler: I plan to use this for the integration tests.

PR #157, opened 2020-03-27
Add logging to common and worker

nikolatzotchev:
This is some more logging on the worker and common, they can me maybe some redundant parts, and I don't want to add to much. If I missed something let me know.

### DIFF
--- a/src/ja/common/proxy/command_handler.py
+++ b/src/ja/common/proxy/command_handler.py
@@ -5,6 +5,9 @@ from ja.common.message.base import Response
 import yaml
 from typing import Dict
 import grp
+import logging
+
+logger = logging.getLogger(__name__)
 
 
 class CommandHandler(ABC):
@@ -87,6 +90,8 @@ class CommandHandler(ABC):
                     type_name=input_dict["type_name"],
                     username=input_dict["username"]
                 )
+                logger.info("handling %s command" % input_dict["type_name"])
+                logger.debug(input_dict["command"])
                 response_string = yaml.dump(response_dict)
                 connection.sendall(response_string.encode())
             finally:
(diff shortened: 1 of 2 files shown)


src/ja/common/proxy/remote.py
```diff
@@ -21,6 +24,7 @@ def __init__(self, socket_path: str, input_stream: TextIO = stdin, output_stream
         @param socket_path: The Unix named socket to write the Command to.
         @param output_stream: The output stream to write the Response to.
         """
+        logger.info("creating socket")
```
ammen99: The remote should not print anything, we parse its stdout to get the response.

src/ja/common/proxy/command_handler.py
```diff
@@ -71,6 +74,7 @@ def main_loop(self) -> None:
         while self._running:
             connection, client_address = named_socket.accept()
             try:
+                logger.info("handling command")
```
ammen99: I think it would make sense to print the command that we receive, not just "handling command". So you could print the type of the received command as info and the command itself as debug.

src/ja/worker/docker.py
```diff
@@ -43,6 +46,8 @@ def worker_uid(self) -> str:
     def add_job(self, job: Job) -> None:
         if job.uid in self._containers_by_job_uid:
             raise ValueError("Job with UID %s already exists." % job.uid)
+        logger.info("adding job: %s" % job.uid)
```
ammen99: I think it would also make sense to print the docker ids, at least on the debug level

src/ja/worker/docker.py
```diff
@@ -56,22 +56,26 @@ def add_job(self, job: Job) -> None:
             mem_limit="%sm" % job.docker_constraints.memory, user=job.owner_id
         )
         self._jobs_by_container_id[container.id] = job
+        logger.debug("container id: " + container.id)
```
ammen99: Also print what we're doing with this container xD

PR #158, opened 2020-03-27
Simple integration tests

JohannesGaessler:
This PR aims to add simple integration tests.

Among others I have found and fixed the following bugs:

- Buggy interaction between deepcopy in the scheduler and sqlalchemy, fixed by overriding deepcopy for jobs.
- Changing the assigned work machine of a job added another copy of the work machine to the database, fixed by looking up a work machine from the database to use as the assigned work machine.

I also made the following misc changes:
- Improved help messages for user client CLI.
- username, email now no longer mandatory CLI arguments.


nikolatzotchev:
` remote_module=WORKER_REMOTE_MODULE % self._uid,` that is probably from the last PR, but why is the remote set to this, when the file in ja-integration is called worker_0.py this brakes when I try to run it(doesn't brake but I get an error message from the logger)

nikolatzotchev:
also I don't now how link it but on line 67 `            config_path=WORKER_CONF_PATH % self._index, socket_path=WORKER_SOCKET_PATH,` shouldn't the socket_path be WORKER_SOCKET_PATH % self._index ?

JohannesGaessler:
@nikolatzotchev Thanks for the input, it should be fixed now.

nikolatzotchev:
How do we check that the tests work, because right now we just see that nothing crashes, do we just read the logs. For example how can we check that the job was successfully executed on the machine (in python).

JohannesGaessler:
I have added assertions to the tests: That there is 1 active job on the server and worker while the job is running and 0 jobs on the server and worker after the job has finished.

nikolatzotchev:
When I was testing, there was a "error" because the job didn't have an uid, that was when the job was finished and the worker sends the message I worked when the uid was set.

JohannesGaessler:
That is fixed now.
The PR for setting job UIDs was merged and this PR was rebased onto master.

nikolatzotchev:
aa ok

PR #159, opened 2020-03-27
set uid

### LINKED ISSUE #136: Assign UIDs to jobs in the AddCommand.execute() method
The current implementation of AddCommand.execute() is wrong, since the job UID is not set before the job is added in the database.

We need to somehow generate a UID, and also add tests for that. The command doesn't have any tests!

M1keReck:
fixes #136

### DIFF
--- a/src/ja/server/database/sql/database.py
+++ b/src/ja/server/database/sql/database.py
@@ -1,7 +1,9 @@
 from copy import deepcopy
 from typing import List, Optional, Callable
 from datetime import datetime
+import time
 from sqlalchemy import create_engine
+from pwd import getpwuid
 
 from ja.common.work_machine import ResourceAllocation
 from ja.common.docker_context import DockerContext, DockerConstraints, MountPoint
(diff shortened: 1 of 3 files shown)


ammen99:
Is this UID then printed back in the user cli? It should be returned in the response.

nikolatzotchev:
@ammen99 bwt travis seems away again. should maybe restart him one more time

M1keReck:
@nikolatzotchev I am actually getting emails from travis, this PR passed all checks

nikolatzotchev:
If all checks pass it looks good to me.

src/ja/server/database/sql/database.py
```diff
@@ -202,11 +203,13 @@ def find_job_by_label(self, label: str) -> List[Job]:
...
         return jobs
 
-    def update_job(self, job: Job) -> None:
+    def update_job(self, job: Job) -> str:
         session = self.scoped()
         old_job_entry: Optional[DatabaseJobEntry] = self.find_job_by_id(job.uid)
         old_job = old_job_entry.job if old_job_entry else None
         if old_job is None:
+            if job.uid is None:
+                job.uid = str(job.owner_id) + str(int(time.time() * 1000))
```
JohannesGaessler: Please add `from pwd import getpwuid` to your imports and replace `str(job.owner_id)` with `getpwuid(job.owner_id).pw_name`.
That way job UIDs will be something like user12345 instead of just a string of numbers.

PR #171, opened 2020-03-28
print results in main

### LINKED ISSUE #169: Jobadder user doesn't output to stdout when command is executed.
As far as I see we don't output the response to the stdout of the user.

M1keReck:
The responses of the use commands should be printed out somewhere
fixes #169


M1keReck:
@nikolatzotchev wdym?

nikolatzotchev:
this just prints the job as a dict and doesn't look very good, but it is fine for now.

PR #172, opened 2020-03-28
Fix bug QueryCommand.

### LINKED ISSUE #170: QueryComand is not workin when one of the fields is None.
this is my mistake when making the `from_dict()` method.

nikolatzotchev:
fixes #170,
basically i was doing stuff wrong should be good now.

### DIFF
--- a/src/ja/common/message/base.py
+++ b/src/ja/common/message/base.py
@@ -59,6 +59,8 @@ class Serializable(ABC):
     @classmethod
     def _get_str_list_from_dict(cls, property_dict: Dict[str, object], key: str, mandatory: bool = True) -> List[str]:
         prop = cls._get_from_dict(property_dict=property_dict, key=key, mandatory=mandatory)
+        if prop is None:
+            return prop
         if mandatory and not isinstance(prop, list):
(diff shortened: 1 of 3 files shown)


src/ja/common/message/base.py
```diff
@@ -59,6 +59,8 @@ def _get_str_from_dict(cls, property_dict: Dict[str, object], key: str, mandator
     @classmethod
     def _get_str_list_from_dict(cls, property_dict: Dict[str, object], key: str, mandatory: bool = True) -> List[str]:
         prop = cls._get_from_dict(property_dict=property_dict, key=key, mandatory=mandatory)
+        if prop is None:
```
ammen99: I somehow think the problem here is that you request mandatory for something which isn't mandatory.

nikolatzotchev: `         uid = cls._get_str_list_from_dict(property_dict=property_dict, key="uid", mandatory=False)`
for example this, the uid list can be None and if you are to remove this if above it doesn't work.

nikolatzotchev: this worked if I use the _get_from_dict but then I have to cast and yes.

PR #173, opened 2020-03-28
Scheduler test

### LINKED ISSUE #161: Write integration tests for scheduling
Write integration tests based on https://github.com/DistributedTaskScheduling/JobAdder/pull/158 that test if scheduling works as intended.

ammen99:
This contains a lot of fixes I found when testing.

I put the file in `ja_integration` because my autocomplete does not work in `test/`, I just realized that even though I had this problem ever since the beginning.

Once the PR is approved, I can move the file to where the other tests are if this is a big problem.

EDIT: forgot to open this as draft, I will be adding a few more tests but I hope I don't need to fix as many issues as I had to up to now :)

Fixes #161

### DIFF
--- a/src/test/integration/test_scheduler.py
+++ b/src/test/integration/test_scheduler.py
@@ -1,0 +1,122 @@
+from ja.common.job import JobStatus, Job
+from ja.server.scheduler.default_policies import DefaultCostFunction
+from test.integration.base import IntegrationTest
+from time import sleep
+from typing import Any, List, Tuple
+
+
+class TestScheduler(IntegrationTest):
+    def __init__(self, *args: Any, **kwargs: Any):
+        super().__init__(*args, **kwargs)
+        self.list_updates: List[Tuple[str, JobStatus]] = []
+
+    def _status_updated(self, job: Job) -> None:
+        self.list_updates.append((job.label, job.status))
+
+    def test_three_jobs(self) -> None:
+        """
+        A test where 3 jobs with different priorities are scheduled, and then a worker is added.
+        """
+        self._clients[0].run(
+            self.get_arg_list_add(num_seconds=1, label="lo", priority="0", threads=4, memory=16 * 1024))
+        self._clients[1].run(
+            self.get_arg_list_add(num_seconds=1, label="me", priority="1", threads=4, memory=16 * 1024))
+        self._clients[2].run(
+            self.get_arg_list_add(num_seconds=1, label="hi", priority="3", threads=4, memory=16 * 1024))
+        self._server._database.set_job_status_callback(self._status_updated)
+        self._add_worker(0)
+        sleep(8)
+        expect: List[Tuple[str, JobStatus]] = [
+            ("hi", JobStatus.RUNNING),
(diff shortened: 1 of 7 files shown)


JohannesGaessler:
>Once the PR is approved, I can move the file to where the other tests are if this is a big problem.

Please do, just to keep things organized.
Also, you didn't adjust .travis.yml to include the tests in ja_integration, did you?

JohannesGaessler:
A rebase onto master would be convenient for reviewing.
Also, please notify us when this PR has all the tests you plan to implement.

ammen99:
I think it should be ready for review now.

I intend to implement cancelling jobs in a separate PR because it will require changes to the database, scheduler and of course a new test.

ammen99:
Btw I moved the two simple test cases from `base.py` to `test_simple.py` because I don't think we want to run them multiple times for each test case.

nikolatzotchev:
2020-03-29-144014_grim
why do I get this one when running the tests.

src/test/integration/test_scheduler.py
```diff
@@ -0,0 +1,122 @@
...
+        self._clients[0].run(
+            self.get_arg_list_add(num_seconds=5, label="lo", priority="0", threads=4, memory=16 * 1024))
+        self._clients[1].run(
+            self.get_arg_list_add(num_seconds=5, label="me", priority="1", threads=2, memory=8 * 1024))
+        self._clients[2].run(
+            self.get_arg_list_add(num_seconds=5, label="hi", priority="3", threads=2, memory=8 * 1024))
+        sleep(1)
+        # Jobs can't have finished so they are all in the schedule
+        distribution = self._server._database.get_current_schedule()
+        worker = sorted([job.assigned_machine.uid for job in distribution])
```
nikolatzotchev:
```suggestion
        workers = sorted([job.assigned_machine.uid for job in distribution])
```

src/test/integration/test_scheduler.py
```diff
@@ -0,0 +1,122 @@
...
+            ("lo", JobStatus.DONE),
+        ]
+        self.assertListEqual(self.list_updates, expect)
+
+    def test_parallel_jobs(self) -> None:
+        """
+        A test where 3 jobs are scheduled on 3 workers
+        """
+        self._add_worker(0)
+        self._add_worker(1)
```
JohannesGaessler: Aren't there just 2 workers?

ammen99: Oh yes, there used to be 3 but I changed it so that we can also test multiple jobs on the same machine. Let me fix that

JohannesGaessler:
>why do I get this one when running the tests.

Probably because our packages are not on your PATH / PYTHONPATH.
You could fix this by doing either of the following:

- Install JobAdder via the install script.
- Make it so that **command** SSH shells connecting to your computer correctly set PYTHONPATH. This can **not** be achieved by editing .bashrc, .profile, etc. Look at .travis.yml for instructions.

ammen99:
@JohannesGaessler It was my fault, I had forgotten to git add the remotes.

PR #176, opened 2020-03-28
Installation guide, misc adjustments

JohannesGaessler:
Fixes https://github.com/DistributedTaskScheduling/JobAdder/issues/166 .

This PR adds an installation guide.
It also does a few related changes:

* Fixed server config database port not actually being used.
* Passwords no longer mandatory properties for login configs.
* Create jobadder system account in travis script.
* Add jobadder user to the docker group in travis script.
* Removed read permissions for server/worker configuration files because they contain passwords.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,0 +1,91 @@
+This file gives a detailed explanation of how to install JobAdder on user clients, the server, or worker clients.
+Note that all components of JobAdder were designed to be used exclusively on Linux.
+Furthermore, the server and the worker clients are assumed to be using central authentication (e.g. LDAP).
+
+## Requirements
+The following software is required to run JobAdder:
+
+- Python, 3.7 or higher (user client, server, and worker clients)
+- PostgreSQL (server)
+- Docker, recent version (worker clients)
+
+## Installation Notes (Linux)
+Detailed step-by-step instructions to install JobAdder.
+
+### 1. Install JobAdder
+The git repository can be cloned by running:
+    
+    git clone https://github.com/DistributedTaskScheduling/JobAdder
+
+After that, execute the *install.sh* script located in the git repository's root directory with superuser privileges.
+This will install JobAdder system-wide through pip.
+
+### 2. Post-Installation Instructions
+Additional steps to take after the installation of JobAdder.
+#### 2.1 Post-Installation Instructions (Server)
+Additional steps to take on the server.
+##### 2.1.1 Create the jobadder User and Group
+Add the *jobadder* user and the corresponding group by running:
+
+    sudo useradd --system jobadder
+
+The jobadder user is used by the daemons running on the server and the worker clients.
+Users in the jobadder group are granted administrative privileges in the context of JobAdder: 
+they can manage worker clients and the jobs of any user.
+
+The jobadder user needs to be added to the docker group to run jobs:
+
+    sudo usermod -a -G docker jobadder
+
+##### 2.1.2 Configure PostgreSQL
+Create a PostgreSQL user called *jobadder*:
+
+    sudo -u postgres createuser --pwprompt jobadder
+
+Create a PostgreSQL database called *jobadder*:
+
+    sudo -u postgres createdb --owner=jobadder jobadder
+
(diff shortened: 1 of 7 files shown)


INSTALL.md
```diff
@@ -0,0 +1,89 @@
+This file gives a detailed explanation of how to install JobAdder on user clients, the server, or worker clients.
+Note that all components of JobAdder were designed to be used exclusively on Linux.
+Furthermore, the server and the worker clients are assumed to be using central authentication (e.g. LDAP).
+
+## Requirements
+The following software is required to run JobAdder:
+
+- Python 3.7 or higher (user client, server, and worker clients)
+- PostgreSQL (server)
+- Docker (worker clients)
```
ammen99: We need a version, at least say "recent version" (some distros do have very outdated docker)

INSTALL.md
```diff
@@ -0,0 +1,89 @@
...
+
+The jobadder user is used by the daemons running on the server and the worker clients.
+Users in the jobadder group are granted administrative privileges in the context of JobAdder: 
+they can manage worker clients and the jobs of any user.
+
+The jobadder user needs to be added to the docker group to run jobs:
+
+    sudo usermod -a -G docker jobadder
+
+##### 2.1.2 Configure PostgreSQL
```
ammen99: The user may need to adjust PostgreSQL authentication permissions, depending on the distribution's default settings. Something like `https://github.com/DistributedTaskScheduling/documentation/pull/88/files#diff-39e2ddcedb8cf74ada971b3371c4206fR20-R26` (maybe only as a tip, but it would save time for somebody unfamiliar with the database)

src/ja/server/config.py
```diff
@@ -73,7 +73,7 @@ def from_dict(cls, property_dict: Dict[str, object]) -> "LoginConfig":
         host = cls._get_str_from_dict(property_dict=property_dict, key="host")
         port = cls._get_int_from_dict(property_dict=property_dict, key="port", mandatory=False)
         username = cls._get_str_from_dict(property_dict=property_dict, key="username")
-        password = cls._get_str_from_dict(property_dict=property_dict, key="password")
+        password = cls._get_str_from_dict(property_dict=property_dict, key="password", mandatory=False)
```
ammen99: When is password not needed?

JohannesGaessler: When I changed the default in the config from test123 to None.
The change to the mandatory keyword makes it so that the server doesn't immediately crash in that case.

ammen99: But without a password, it probably won't work at all? Or does it work?

JohannesGaessler: I forgot that we don't have unit tests for the actual server configuration files; it does not work.
I made passwords mandatory again.

INSTALL.md
```diff
@@ -0,0 +1,90 @@
...
+Create a PostgreSQL user called *jobadder*:
+
+    sudo -u postgres createuser --pwprompt jobadder
+
+Create a PostgreSQL database called *jobadder*:
+
+    sudo -u postgres createdb --owner=jobadder jobadder
+
+Make sure to configure the [authentication method](https://www.postgresql.org/docs/12/auth-methods.html) of your PostgreSQL installation.
```
ammen99: Should also say which auth method we need.

JohannesGaessler:
>lgtm, I think it would also bw nice to say we want md5/password auth.

Authentication for what?

ammen99:
> > lgtm, I think it would also bw nice to say we want md5/password auth.
> 
> Authentication for what?

Password authentication for the database. I can imagine a person not familiar with this file will have difficulties, just as I had, so that's why I think we should say that the admin needs to enable md5 authentication method.

JohannesGaessler:
Just to make sure there isn't a misunderstanding:
I added a line below the link to authentication methods that mentions that JobAdder was designed with password authentication in mind.
Is this not sufficient?

ammen99:
I had missed the second line, sorry.

ammen99:
If you squash your commits, it is ready for merge.

ammen99:
@JohannesGaessler By the way, you can also merge pull requests :)

PR #177, opened 2020-03-29
Add test for quering job.

### LINKED ISSUE #164: Integration test for querying jobs

nikolatzotchev:
Add 2 jobs and runs them, after that query one of them, compare stdout to check if job is correct.
Not sure if that is the best way for checking stdou but it works.
fixes #164


nikolatzotchev:
This test is enough right, maybe I will add one more where the query command returns more then 1 job, but if the commands have unnittests this should suffice.

src/test/integration/test_quering.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def test_query_one_job(self) -> None:
+        client = self._clients[0]
+        label = "not_a_bitcoin_miner"
+        client.run(self.get_arg_list_add(num_seconds=2, label=label))
+        client.run(self.get_arg_list_add(num_seconds=2, label="123"))
+        sleep(4)
+        with patch("sys.stdout", new=StringIO()) as fakeOutput:
+            client.run(self._get_arg_list_query())
+            active_job_entries = self._server._database.find_job_by_label("not_a_bitcoin_miner")[0]
```
JohannesGaessler: You probably forgot to rename *active_job_entries* after you copy-pasted but it's not like it affects functionality.

nikolatzotchev: ops.

PR #180, opened 2020-03-29
More fixes

ammen99:
Fixes two bugs I encountered when running the tests for the cancel command. I have them running locally, and will submit the pull request when we get this PR, #179 and #173 since otherwise I'll have to keep rebasing.

### DIFF
--- a/src/ja/server/database/types/work_machine.py
+++ b/src/ja/server/database/types/work_machine.py
@@ -14,17 +14,21 @@ class WorkMachineResources(Serializable):
     how much of its resources are currently used.
     """
 
-    def __init__(self, total_resources: ResourceAllocation):
+    def __init__(self, total_resources: ResourceAllocation, free_resources: ResourceAllocation = None):
         """!
         Initialize the RunningWorkMachineResources object. In the beginning,
         all resources are marked as free.
         @param total_resources The total amount of available resources.
+        @param free_resources The amount of free resources, if None, then all resources are free.
         """
         if total_resources.is_negative():
             raise ValueError("None of the resources can be negative!")
 
         self._total_resources = ResourceAllocation.from_dict(total_resources.to_dict())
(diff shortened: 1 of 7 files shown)


src/ja/server/database/types/work_machine.py
```diff
@@ -134,6 +134,16 @@ def __eq__(self, o: object) -> bool:
                 and self.resources == o.resources and self.state == o.state
         return False
 
+    def __deepcopy__(self, memodict: Dict[str, object] = {}) -> "WorkMachine":
+        res_copy: WorkMachineResources = None
+        ssh_conf: SSHConfig = None
+        if self.resources:
+            res_copy = WorkMachineResources.from_dict(self.resources.to_dict())
+            res_copy.allocate(self.resources.total_resources - self.resources.free_resources)
```
JohannesGaessler: I assume that not including the above line causes a bug because resource allocation is not tracked when (de)serializing WorkMachineResources; therefore from_dict(to_dict(...)) would not work for WorkMachine.
Is this a problem for us?
In other words, are there instances where we need to transfer a WorkMachineResources object from one component to another that needs to transfer information about currently allocated resources?
(It might have been a cleaner fix to adjust to_dict and from_dict of WorkMachineResources)

JohannesGaessler: Sure, I just want to make sure that we are not overlooking instances where we would need the more elaborate fix anyways.
Like, if we had implemented querying the workload of workers we would probably need the more elaborate fix.
I can't think of any scenario where we would need the more elaborate fix, but someone else might.

ammen99: Okay I'll try to do it the other way and see how it goes, if there are a lot of broken tests I won't fix them.

ammen99: Went easier than I expected ...

PR #186, opened 2020-03-30
dispatcher: sort jobs before dispatching

### LINKED ISSUE #174: Dispatcher: first cancel/pause jobs, then start new ones
Otherwise, the worker could temporarily start thrashing since both an old and a new preemptive task could run concurrently.

If we are left with time, we could do that since it is relatively easy, just need to sort the distribution so that cancel/pause is before the others.

ammen99:
This makes sure we have no races between cancelled and newly started jobs, as well as that we first pause before starting jobs on a machine.

Fixes #174


ammen99:
> I'll revert my commit...

I didn't know when you'll be back online I already rebased your branch on top of this one and fixed some review comments, check out `timeout-2`

PR #193, opened 2020-03-31
Loosened integration test timings to account for low-power CPUs

JohannesGaessler:
I installed JobAdder on a 1.2 GHz ARM SBC and found that several integration tests were failing.
This was due to the current timings being too tight for very slow systems.
This PR loosens those timings so that the integration tests succeed on slow system.
The looser timings extend the time required to complete the tests by 27 seconds.

### DIFF
--- a/src/test/integration/test_cancel.py
+++ b/src/test/integration/test_cancel.py
@@ -15,9 +15,9 @@ class TestJobCancel(IntegrationTest):
         Run one job, queue another and cancel both.
         """
         self._clients[0].run(
-            self.get_arg_list_add(num_seconds=5, label="lo", priority="0", threads=4, memory=16 * 1024))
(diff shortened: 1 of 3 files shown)


JohannesGaessler:
I disagree.
I think that ensuring the tests work correctly on all (or at least most) platforms is much more important than saving a few seconds in testing.

ammen99:
> I disagree.
> I think that ensuring the tests work correctly on all (or at least most) platforms is much more important than saving a few seconds in testing.

I agree with your argument when it comes to platforms which are our target hardware, but your NanoPi is not one of those :)

Plus, tests are for developers, so it should work on our machines. Faster tests also let you develop code & fix bugs faster. This does not apply anymore, since we are done with both implementation and QA.

PR #194, opened 2020-03-31
Removed hard-coded pip prefix

JohannesGaessler:
Currently the install script uses /usr as the hard-coded pip prefix.
However, when I installed JobAdder with this prefix on my SBC my python installation (and pip3) was not able to find it afterwards.
The operating system in use was a derivative of Ubuntu 16.04.

This PR removes the hard-coded prefix for the pip installation and instead lets pip decide where to install JobAdder.
Since the script is executed as root this will still be a system-wide installation.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -7,12 +7,11 @@ if [[ $EUID -ne 0 ]]; then
    exit 1
 fi
 
-PREFIX=/usr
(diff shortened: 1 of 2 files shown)


JohannesGaessler:
One of the unittests that should be completely unrelated to the pip change was failing.
I suspect that the test had a race condition due to strict timings.
Setting loose timings seems to have fixed the test.

PR #197, opened 2020-03-31
Fixed config file ownership

JohannesGaessler:
On master the config files for server and worker are owned by root:root with 600 permissions.
However, since the server and worker daemons are executed as the user jobadder this means that the daemons don't have permission to read the config files.
This PR changes config file ownership to jobadder:jobadder with 600 permissions; the daemons can read and write config files but users in the jobadder group and regular users still do not have permission.


JohannesGaessler:
You're right.
I re-installed JobAdder on a machine where the account was already created so I forgot.

PR #198, opened 2020-04-03
improved query responses

### LINKED ISSUE #190: Improve query responses
Are there any suggestions?

M1keReck:
This is what an output for one job looks like (if verbosity is not set to 2):
`Job uid: 0, label: lab, status: JobStatus.QUEUED, assigned work machine uid: None`

fixes #190

### DIFF
--- a/src/ja/user/cli.py
+++ b/src/ja/user/cli.py
@@ -92,7 +92,8 @@ class UserClientCLIHandler:
 
         parser: ArgumentParser = ArgumentParser("JobAdder: add, cancel or a query a job.")
         # shared options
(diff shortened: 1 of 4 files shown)


ammen99:
I think something like http://ix.io/2gCp would be sufficient, in this way, we don't have to use external library while still providing relatively sane output.

M1keReck:
lol I was just about to push the tabulate implementation, but yeah that's better

ammen99:
> lol I was just about to push the tabulate implementation, but yeah that's better

Oh sorry then.

Issue #168, opened 2020-03-28
Admin jobs are assigned uids `rootXXX`


ammen99:
The issue is that for these, we set the owner to 0 since this is the effective user value indicating admin ownership. To fix, the owner id should be set in the user CLI.

This is low priority though.

PR #201, opened 2020-04-04
Fixed effective user being root for admin users

JohannesGaessler:
Fixes https://github.com/DistributedTaskScheduling/JobAdder/issues/168 .

On master the effective user for users in the jobadder group is set to 0, i.e. root.
This is **not** just a cosmetic issue.
Jobs submitted by users in the jobadder group were run as root, effectively granting root access to all users in the jobadder group.
This PR revises how a user being admin is checked to prevent this.

### DIFF
--- a/src/ja/user/message/add.py
+++ b/src/ja/user/message/add.py
@@ -37,7 +37,7 @@ class AddCommand(UserServerCommand):
         if job.owner_id == -1:
             job.owner_id = self.effective_user
 
-        if self.effective_user != 0 and self.effective_user != job.owner_id:
+        if not self.effective_user_is_root and self.effective_user != job.owner_id:
             return Response(result_string="Cannot submit jobs for other users.", is_success=False)
 
         db_job_id: DatabaseJobEntry = database.find_job_by_id(job.uid)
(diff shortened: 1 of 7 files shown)


JohannesGaessler:
I think we should just disallow submitting jobs for other users (or maybe just allow root to do so?).
I think it is not a very useful feature anyways and security is more important.

src/ja/user/message/base.py
```diff
@@ -17,6 +20,23 @@ def effective_user(self) -> int:
...
+    @property
+    def admin_group(self) -> str:
+        return self._admin_group
+
+    @admin_group.setter
+    def admin_group(self, admin_group: str) -> None:
+        self._admin_group = admin_group
+
+    @property
+    def effective_user_is_admin(self) -> bool:
```
ammen99, line 32: So, my question is, why duplicate the code here and in the command handler? Can't we just set `user_is_admin` instead of giving the admin group?

src/ja/user/message/add.py
```diff
@@ -37,7 +37,7 @@ def execute(self, database: ServerDatabase) -> Response:
         if job.owner_id == -1:
             job.owner_id = self.effective_user
 
-        if self.effective_user != 0 and self.effective_user != job.owner_id:
+        if not self.effective_user_is_root and self.effective_user != job.owner_id:
```
ammen99, line 40: I see this as just as bad a security hole as before. I can run jobs as any user and get all of their files. After all, we want to protect the user's files.

Didn't we say we will remove submitting jobs as other users completely?

ammen99: Ah I am stupid, I didn't read it carefully enough. It is a good solution I think.

